PR #2, opened 2015-04-01
fix to clone repository without master branch

birmacher:
checked
- clone without --branch
- clone without --branch and running cocoapods & repository validator
- clone with --branch


birmacher:
GitHub pull request support also added to this PR

PR #3, opened 2015-07-31
Fix for destination directory

birmacher:
Destination directory path was ignored

### DIFF
--- a/step.yml
+++ b/step.yml
@@ -45,7 +45,7 @@ inputs:
     is_expand: true
   -
     title: "Clone destination (local) directory path"
-    mapped_to: BITRISE_SOURCE_DIR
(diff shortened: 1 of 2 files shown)


step.yml
```diff
@@ -11,10 +11,10 @@ description: |
 
   Use either the *(raw) SSH* or the *Base64 encoded SSH* key input to specify
   the SSH key. If both specified the *(raw) SSH* key input will be used.
-website: https://github.com/bitrise-io/steps-git-clone
-fork_url: https://github.com/bitrise-io/steps-git-clone
+website: https://github.com/birmacher/steps-git-clone
```
viktorbenei: nop ;)

birmacher: just updated;)

PR #11, opened 2016-03-09
Quote tag and branch names

MisterRager:
Branches with characters like the single-quote (`'`) cause builds to hang indefinitely. This should solve that.

### DIFF
--- a/step.sh
+++ b/step.sh
@@ -15,8 +15,8 @@ fi
 ruby "${THIS_SCRIPT_DIR}/git_clone.rb" \
 	--repo-url="${repository_url}" \
 	--commit-hash="${commit}" \
-	--tag="${tag}" \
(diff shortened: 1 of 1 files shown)


viktorbenei:
Are you sure that there's a problem with this step? We know about an escaping issue, but that's in https://github.com/bitrise-tools/bitrise-machine or in one of the sub-tools it uses to communicate with the build VMs.

We'll fix that this week.

MisterRager:
Thanks - I just know that *something* hangs *somewhere*.

viktorbenei:
You're right, it's actually an escaping issue in one of the core tools used for pushing the parameter into the build VM. We'll fix it ASAP!

MisterRager:
Yay!

PR #13, opened 2016-03-31
clone depth

godrei:


bitrise.yml
```diff
@@ -37,6 +38,7 @@ workflows:
         run_if: true
         inputs:
         - clone_into_dir: repo-dir
+        - clone_depth: $CLONE_DEPTH
```
viktorbenei: input from where?

step.yml
```diff
@@ -44,6 +44,9 @@ inputs:
   - auth_ssh_private_key: "$SSH_RSA_PRIVATE_KEY"
     opts:
       title: "Auth: SSH private key - without a passphrase!"
+  - clone_depth:
+    opts:
+      title: "Limit fetching to the specified number of commits"
```
viktorbenei: add an example - clarification that this has to be a number

step.yml
```diff
@@ -44,6 +44,13 @@ inputs:
...
+      description: |-
+        Limit fetching to the specified number of commits.
+        The depth number should present as a string.
```
viktorbenei: The value should be a decimal number, example: `10`

PR #24, opened 2016-10-15
support for head branch when exposed by pull requests

birmacher:


viktorbenei:
Is this required @birmacher ? Should we update this?

viktorbenei:
we won't use this for now

PR #29, opened 2017-03-01
Retry checkout on fail with feching tags

trapacska:


main.go
```diff
@@ -174,7 +174,21 @@ func main() {
 			}
 		}
 
-		if err := git.Checkout(); err != nil {
+		if err := retry.Times(1).Wait(waitTime).Try(func(attempt uint) error {
```
godrei: use the global `retryCount` instead of `1`

Issue #30, opened 2017-04-11
Bitbucket / diff based PR: step fails if there's 0 diff (nothing changed in the PR)


viktorbenei:
```
+------------------------------------------------------------------------------+
| (2) git-clone@3.4.2                                                          |
+------------------------------------------------------------------------------+
| id: git-clone                                                                |
| version: 3.4.2                                                               |
| collection: https://github.com/bitrise-io/bitrise-steplib.git                |
| toolkit: go                                                                  |
| time: 2017-04-10T02:45:11-07:00                                              |
+------------------------------------------------------------------------------+
|                                                                              |
INFO[02:45:11]  * [OK] Step dependency (go) installed, available. 

Git Clone Configs:
- CloneIntoDir: /Users/vagrant/git
- RepositoryURL: git@bitbucket.org:REDACTED/REDACTED2.git
Git Checkout Configs:
- Commit: 8d7fd0e400c9
- Tag: 
- Branch: test/no_change_pull_request
- CloneDepth: 
Git Pull Request Configs:
- PullRequestURI: git@bitbucket.org:REDACTED/REDACTED2.git
- PullRequestID: 262
- BranchDest: release/1.4.1
- PullRequestMergeBranch: 
- ResetRepository: No
Bitrise Build Configs:
- BuildURL: https://www.bitrise.io/build/REDACTED3
- BuildAPIToken: z_Z3dqyFd1ttLzWnB4R8Ew

Git clone repository
=> git "init"
=> git "remote" "add" "origin" "git@bitbucket.org:REDACTED/REDACTED2.git"
=> git "fetch"
=> git "checkout" "release/1.4.1"
=> git "apply" "/var/folders/90/5stft2v13fb_m_gv3c8x9nwc0000gn/T/262.diff240882616"
0 attempt failed:
error: unrecognized input

Retrying...
=> git "apply" "/var/folders/90/5stft2v13fb_m_gv3c8x9nwc0000gn/T/262.diff240882616"
1 attempt failed:
error: unrecognized input

Retrying...
=> git "apply" "/var/folders/90/5stft2v13fb_m_gv3c8x9nwc0000gn/T/262.diff240882616"
2 attempt failed:
error: unrecognized input

Failed, error: error: unrecognized input

|                                                                              |
+---+---------------------------------------------------------------+----------+
| x | git-clone@3.4.2 (exit code: 1)                                | 37 sec   |
+---+---------------------------------------------------------------+----------+
[...]

PR #31, opened 2017-04-11
#30 fixed, check if diff file is existing and has content

trapacska:


godrei:
maybe godeps-update is missing?

`cannot find package "github.com/bitrise-io/go-utils/fileutil"`

PR #41, opened 2017-10-07
Add commit count output

### LINKED ISSUE #23: Feature request: add commit count as output variable from this step
In our project we would like to use the commit count instead of the build number for versioning. 
It would be useful for this step to output a variable with the commit count.
`
GIT_CLONE_COMMIT_COUNT = git rev-list --count HEAD`

ninjaprox:
Implement the feature request in #23 

closes #23


erosdome:
Thanks for the pull request :) @godrei could you take a look at it?

godrei:
Hi @ninjaprox ,

Count will only work if no "depth" is set. This will be highlighted in the related output's description.

If "clone depth" is set then not the whole history will be available, and count will NOT fail but will be the clone depth (quite useless).

PR #42, opened 2017-10-16
Adds git apply arguments for PR cloning

nhammond101:
ability to pass arguments to the git patch apply command ie `git apply --cached <patch>`

### DIFF
--- a/gitutil/gitutil.go
+++ b/gitutil/gitutil.go
@@ -32,7 +32,7 @@ type PullRequestHelper struct {
 	pullRequestBranch        string
 	pullRequestMergeBranch   string
 	pullRequestDiffPath      string
-	PullRequestPatchArgs     string
+	PullRequestPatchArgs     []string
 }
 
 // Helper ...
@@ -148,11 +148,11 @@ func (helper *Helper) ConfigureCheckoutWithPullRequestURI(pullRequestID, pullReq
 		pullRequestID:            pullRequestID,
 		pullRequestRepositoryURI: pullRequestURI,
 		pullRequestBranch:        pullRequestBranch,
-		PullRequestPatchArgs:     "",
+		PullRequestPatchArgs:     []string{},
 	}
 
 	if patchArgs != ""{
-		helper.pullRequestHelper.PullRequestPatchArgs = patchArgs
+		helper.pullRequestHelper.PullRequestPatchArgs = strings.Split(patchArgs," ")
 	}
 
 	helper.cloneDepth = cloneDepth
@@ -164,11 +164,11 @@ func (helper *Helper) ConfigureCheckoutWithPullRequestID(pullRequestID, pullRequ
 	helper.pullRequestHelper = PullRequestHelper{
 		pullRequestID:          pullRequestID,
 		pullRequestMergeBranch: pullRequestMergeBranch,
-		PullRequestPatchArgs:   "",
+		PullRequestPatchArgs:   []string{},
 	}
 
 	if patchArgs != ""{
-		helper.pullRequestHelper.PullRequestPatchArgs = patchArgs
+		helper.pullRequestHelper.PullRequestPatchArgs = strings.Split(patchArgs," ")
 	}
 
 	helper.cloneDepth = cloneDepth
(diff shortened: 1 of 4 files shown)


viktorbenei:
Hi @nhammond101 👋  :)

First of all, thanks for the PR!

I'm just curious, what would be the use case for this input?
From first look it seems this would be better to do in a separate step (maybe with a script) in your workflow, as `git apply` is only used as a fallback, in case the git hosting service does provide a diff but not a pre-merged branch for the Pull Request. So it's quite rare that this option is used at all, I believe it's only used for Bitbucket repos right now, as Bitbucket is the only one which does provide diff for PRs but not a pre-merge branch.

nhammond101:
hi @viktorbenei, we use bitbucket and submodules for some of the apps.  we noticed that for PR triggered workflows, that if the submodule commit hash had been updated in the source branch that it wasn't being updated in the destination branch.  Therefore, when the submodules were subsequently updated, it wasn't pulling the correct submodule commit, failing the build if the updated submodule was required for compilation, etc.  

tbh, git submodules is a new subject for the team, so googling, stackoverflowing(? ;) ) led me to try adding the `--cached` and `--index` flags to the `git apply` command, with successful results.  I hadn't realised that `git apply` was a fallback (i'm not familiar with golang either 😊 ), but I tried to keep the minimal amount of changes to achieve the adding of the flags.

i had issues getting the tests to run locally but assumed they were my environment related, so i do apologise if i've missed something simple with respect to the failing ci build.

Incidentally (and pretty off topic), how does one setup an environment properly for golang dev?  i managed for another repo, but the same setup failed for this one.  I used [gvm](https://github.com/moovweb/gvm/).

viktorbenei:
@nhammond101 for local debug: https://github.com/bitrise-steplib/steps-virtual-device-testing-for-android/issues/3#issuecomment-337210349

For the PR: as the git apply with a diff is only used for Bitbucket I don't think we should add this as a new input. Instead we could add these params as constant arguments to `git apply`.

But before that, what kind of issue did this solve? And which flags exactly?
Would it be reasonable to have those flags for every `git apply` call? Or are those special flags for some use case only and should be treated as optional?

nhammond101:
awesome, i've got the local build working better.  it's failing on the linting step.  I re-cloned the repo and it happens on a clean clone too.

**Issue**
master branch has a submodule (module1) with a commit hash of `b3588e4a`, and a feature branch commit hash of `6a81fb11`.  When the PR was raised to merge the feature into master, i'd expect the module1's commit hash to become `6a81fb11`, however running `git submodule status` after the `git "submodule" "update" "--init" "--recursive"` step shows the `b3588e4a` instead.

**Solution(?)**
Adding `--index` to the `git "apply" <patch>` command updates the module1's hash to the correct one.  

I also tested the `--cached` flag and it worked as long as the submodule already existed in the master branch otherwise it's fails with `Failed, error: fatal: No url found for submodule path 'submodules/module2' in .gitmodules`.

I think the `--index` could be added by default, but tbh I don't know understand the command well enough to confirm no adverse side effects for other users.  From what I do understand, as it would be running on a remote CI, the git index can be changed for the build as there's no need to worry about after effects to the local environment.

viktorbenei:
Linter fix is coming: https://github.com/bitrise-io/steps-git-clone/pull/43

For `git apply` - I'll check those flags ASAP!

viktorbenei:
linter fix merged - please upgrade your for and the linter issue should go away :)

nhammond101:
@viktorbenei thanks for that.  local tests pass now :)

viktorbenei:
@nhammond101 I did an update here on the GitHub UI ⬆️

nhammond101:
@viktorbenei any progress?

viktorbenei:
We did not have enough time yet to research the implications of these additional flags @nhammond101 

If these flags would help and not cause any issues we'd add these as fixed/non optional flags, so that it's not required to change anything in the git clone step as a user (no related input).

If you have some time can you research & summarize the implication of these flags? Would help a lot!

nhammond101:
@viktorbenei no worries.  i'll take a look and figure out what exactly they do

viktorbenei:
Thanks @nhammond101 , we'll also do some digging early next week ;)

kdobmayer:
Hi @nhammond101,

Thanks for the PR, we could reproduce the issue with bitbucket and submodules.

The `--cached` flag apply a patch without updating the working tree, the `--index` flag also causes
the index file to be updated and make sure the patch is applicable (it checks if the working tree is up-to-date). If `--index` is not specified, then the submodule commits in the patch are ignored and only the absence or presence of the corresponding subdirectory is checked and (if possible) updated.

To fix this issue there is no need for new input, we   can add `--index` as constant option to `git apply`.

kdobmayer:
The fix will be included in the next release. You can check the commit: https://github.com/bitrise-io/steps-git-clone/pull/44/commits/d3e45dd02c38a5500903f5f3c46230d835c9db9b

PR #52, opened 2017-11-27
Revision

kdobmayer:


main.go
```diff
@@ -102,308 +124,286 @@ func exportEnvironmentWithEnvman(keyStr, valueStr string) error {
...
+}
 
-	if err := configs.validate(); err != nil {
-		log.Errorf("Issue with input: %s", err)
-		os.Exit(1)
-	}
-	fmt.Println()
-	// ---
+func runWithOutput(c *command.Model) (string, error) {
+	//log.Infof(c.PrintableCommandArgs())
```
trapacska: please remove unused code-comment

main.go
```diff
@@ -102,308 +124,286 @@ func exportEnvironmentWithEnvman(keyStr, valueStr string) error {
...
-				log.Errorf("Failed, error: %s", err)
-				os.Exit(1)
-			}
+func runWithRetry(f func() *command.Model) error {
+	return retry.Times(retryCount).Wait(waitTime).Try(func(attempt uint) error {
+		if attempt > 0 {
+			log.Warnf("Retrying...")
+		}
+		err := run(f())
+		//err := run(cmd)
```
trapacska: please remove unused code-comment

main.go
```diff
@@ -38,6 +48,19 @@ type ConfigsModel struct {
 	ManualMerge      string
 }
 
+func init() {
```
godrei: move the `ConfigsModel` initializer (`createConfigsModelFromEnvs`) and it's instance functions (`print`, `validate`) here.

main.go
```diff
@@ -102,308 +124,285 @@ func exportEnvironmentWithEnvman(keyStr, valueStr string) error {
...
-// --- Main
-// -----------------------
-func main() {
-	//
-	// Validate options
-	configs := createConfigsModelFromEnvs()
-
-	fmt.Println()
-	configs.print()
+func run(c *command.Model) error {
```
godrei: move the git related commands to a separate file for cleaner code (`SOURCE_DIR/git.go`)

main.go
godrei: what's the point of init function? why you do not do it's functionality in main.go?

main.go
```diff
@@ -102,308 +124,285 @@ func exportEnvironmentWithEnvman(keyStr, valueStr string) error {
...
+	return c.SetStdout(os.Stdout).SetStderr(os.Stderr).Run()
+}
 
-	if err := configs.validate(); err != nil {
-		log.Errorf("Issue with input: %s", err)
-		os.Exit(1)
-	}
-	fmt.Println()
-	// ---
+func runWithOutput(c *command.Model) (string, error) {
```
godrei: runWithOutput -> runForOutput/runAndReturnOutput

main.go
```diff
@@ -102,308 +124,285 @@ func exportEnvironmentWithEnvman(keyStr, valueStr string) error {
...
+	}
 
-		if commitHash, err := git.LogCommitHash(); err != nil {
-			log.Errorf("Git log failed, error: %s", err)
-			os.Exit(1)
-		} else {
-			log.Printf("=> GIT_CLONE_COMMIT_HASH")
-			log.Printf("   value: %s", commitHash)
-			fmt.Println()
+	if err := run(Git.Init()); err != nil {
```
godrei: would not it fail if repo already initialized? (like in case: `originPresent && configs.ResetRepository != "yes"`)

main.go
```diff
@@ -102,308 +124,285 @@ func exportEnvironmentWithEnvman(keyStr, valueStr string) error {
...
+				log.Warnf("Using Pull Request branch...")
+				branch := strings.TrimSuffix(configs.PullRequestMergeBranch, "/merge")
+				if err := run(Git.Fetch("origin", configs.PullRequestMergeBranch+":"+branch)); err != nil {
+					fail("Fetch Pull Request branch failed (%s), error: %v",
+						configs.PullRequestMergeBranch, err)
+				}
+				if err := run(Git.Checkout(branch)); err != nil {
+					fail("Checkout failed (%s), error: %v", configs.BranchDest, err)
+				}
+			} else {
```
godrei: } else { -> `} else if  diffFilePath != "" {` ?

main.go
```diff
@@ -102,308 +124,285 @@ func exportEnvironmentWithEnvman(keyStr, valueStr string) error {
...
-			log.Errorf("Git log failed, error: %s", err)
-			os.Exit(1)
-		} else {
-			log.Printf("=> GIT_CLONE_COMMIT_MESSAGE_BODY")
-			log.Printf("   value: %s", commitMessageBody)
-			fmt.Println()
-
-			if err := exportEnvironmentWithEnvman("GIT_CLONE_COMMIT_MESSAGE_BODY", commitMessageBody); err != nil {
-				log.Warnf("envman export failed, error: %s", err)
+	if isPR() {
```
godrei: the 3 case if pr:

- configs.ManualMerge == "no" && configs.PullRequestMergeBranch != ""
- configs.ManualMerge == "no" && hasDiffFile
- others (nit github pr nor bitbucket -> configs.ManualMerge does not matter)

main.go
```diff
@@ -102,308 +124,285 @@ func exportEnvironmentWithEnvman(keyStr, valueStr string) error {
...
-			os.Exit(1)
-		} else {
-			log.Printf("=> GIT_CLONE_COMMIT_COMMITER_NAME")
-			log.Printf("   value: %s", commitCommiterName)
-			fmt.Println()
-
-			if err := exportEnvironmentWithEnvman("GIT_CLONE_COMMIT_COMMITER_NAME", commitCommiterName); err != nil {
-				log.Warnf("envman export failed, error: %s", err)
+			if err := runWithRetry(func() *command.Model {
+				return Git.Fetch("--unshallow")
```
godrei: should not we try to unshallow in every casese where fetch fails?

git.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+
+	return run(Git.SubmoduleForeach(Git.Clean("-x", "-d", "-f")))
+}
+
+func isPR() bool {
+	return configs.PullRequestURI != "" || configs.PullRequestID != "" || configs.PullRequestMergeBranch != ""
+}
+
+func setCheckoutArg() string {
```
godrei: why set?

git.go
```diff
@@ -0,0 +1,242 @@
...
+	return diffFile.Name(), nil
+}
+
+func run(c *command.Model) error {
+	log.Infof(c.PrintableCommandArgs())
+	return c.SetStdout(os.Stdout).SetStderr(os.Stderr).Run()
+}
+
+func runForOutput(c *command.Model) (string, error) {
+	return c.RunAndReturnTrimmedCombinedOutput()
```
viktorbenei: shouldn't this `log` as well? similar to `run()`

config.go
```diff
@@ -0,0 +1,86 @@
...
+// ConfigsModel ...
+type ConfigsModel struct {
+	CloneIntoDir  string
+	RepositoryURL string
+	Commit        string
+	Tag           string
+	Branch        string
+	CloneDepth    string
+
+	PullRequestURI         string
```
viktorbenei: I'd rename this to `PullRequestRepositoryURL` or rather `PullRequestRepositoryCloneURL`, as `PullRequestURI` sounds like the URL of the PR page (something like e.g. https://github.com/bitrise-io/steps-git-clone/pull/52)

git.go
```diff
@@ -0,0 +1,242 @@
...
+			log.Warnf("Attempt %d failed:", attempt+1)
+			fmt.Println(err.Error())
+		}
+
+		return err
+	})
+}
+
+func isFork() bool {
+	return configs.RepositoryURL != configs.PullRequestURI
```
viktorbenei: `configs.PullRequestURI` might be empty - if that's the case this won't work

git.go
```diff
@@ -0,0 +1,242 @@
...
+}
+
+func isPrivate() bool {
+	return strings.HasPrefix(configs.RepositoryURL, "git")
```
viktorbenei: shouldn't this check the PR Clone URL (the PR source), instead of the **target** URL?

main.go
```diff
@@ -102,308 +59,86 @@ func exportEnvironmentWithEnvman(keyStr, valueStr string) error {
...
-	// Validate options
-	configs := createConfigsModelFromEnvs()
-
-	fmt.Println()
-	configs.print()
-
-	if err := configs.validate(); err != nil {
-		log.Errorf("Issue with input: %s", err)
-		os.Exit(1)
+	if err := initConfig(); err != nil {
```
viktorbenei: Instead of using a `fail` func or similar in the `main`, I'd suggest to rewrite main to this pattern:

```
func main() {
	if err := doMain(); err != nil {
		log.Errorf("ERROR: %+v", err)
		os.Exit(1)
	}
	log.Donef("\nSuccess")
}
```

Where the actual implementation is in "`doMain`", which returns error (should NOT os.Exit), and the `main()` is only responsible for proper, unified logging of the error.

config.go
```diff
@@ -0,0 +1,86 @@
...
+	PullRequestID          string
+	BranchDest             string
+	PullRequestMergeBranch string
+	ResetRepository        string
```
viktorbenei: should be `bool`

config.go
```diff
@@ -0,0 +1,86 @@
...
+		Commit:        os.Getenv("commit"),
+		Tag:           os.Getenv("tag"),
+		Branch:        os.Getenv("branch"),
+		CloneDepth:    os.Getenv("clone_depth"),
+
+		PullRequestURI:         os.Getenv("pull_request_repository_url"),
+		PullRequestID:          os.Getenv("pull_request_id"),
+		BranchDest:             os.Getenv("branch_dest"),
+		PullRequestMergeBranch: os.Getenv("pull_request_merge_branch"),
+		ResetRepository:        os.Getenv("reset_repository"),
```
viktorbenei: use `bool` instead, like:

```
ResetRepository:        os.Getenv("reset_repository") == "yes",
```

config.go
```diff
@@ -0,0 +1,86 @@
...
+		ManualMerge:            os.Getenv("manual_merge"),
```
viktorbenei: same here, use `bool`

config.go
```diff
@@ -0,0 +1,86 @@
...
+
+		BuildURL:         os.Getenv("build_url"),
+		BuildAPIToken:    os.Getenv("build_api_token"),
+		UpdateSubmodules: os.Getenv("update_submodules"),
```
viktorbenei: use `bool`

git.go
```diff
@@ -0,0 +1,244 @@
...
+func getDiffFile() (string, error) {
+	url := fmt.Sprintf("%s/diff.txt?api_token=%s", config.BuildURL, config.BuildAPIToken)
+	resp, err := http.Get(url)
+	if err != nil {
+		return "", err
+	} else if resp.StatusCode != 200 {
+		return "", fmt.Errorf("Can't download diff file, HTTP status code: %d", resp.StatusCode)
+	}
+	defer func() {
+		if err := resp.Body.Close(); err != nil {
```
viktorbenei: this should be right after `if err != nil` -> you have to close the body in every case if there's no `err`, not just when you read from the body

config.go
```diff
@@ -0,0 +1,101 @@
...
+		PRRepositoryCloneURL: os.Getenv("pull_request_repository_url"),
+		PRID:                 os.Getenv("pull_request_id"),
+		BranchDest:           os.Getenv("branch_dest"),
+		PRMergeBranch:        os.Getenv("pull_request_merge_branch"),
+		resetRepository:      os.Getenv("reset_repository"),
+		manualMerge:          os.Getenv("manual_merge"),
+
+		BuildURL:         os.Getenv("build_url"),
+		BuildAPIToken:    os.Getenv("build_api_token"),
+		updateSubmodules: os.Getenv("update_submodules"),
```
viktorbenei: I'd still rather go for `bool`s for these configs

git.go
```diff
@@ -0,0 +1,244 @@
...
+		return err
+	})
+}
+
+func isFork() bool {
+	return config.PRRepositoryCloneURL != "" &&
+		config.RepositoryURL != config.PRRepositoryCloneURL
+}
+
+func isPrivate() bool {
```
viktorbenei: rename to `isPRSourceRepoPrivate` or similar; the current name does not explain what's private or not

git.go
```diff
@@ -0,0 +1,244 @@
...
+	}); err != nil {
+		return fmt.Errorf("Fetch failed, error: %v", err)
+	}
+
+	if err := run(Git.Checkout(config.BranchDest)); err != nil {
+		return fmt.Errorf("checkout failed (%s), error: %v", config.BranchDest, err)
+	}
+
+	if isFork() {
+		if err := run(Git.RemoteAdd("upstream", config.PRRepositoryCloneURL)); err != nil {
```
viktorbenei: we shouldn't call this "upstream" as "upstream" is the *target*, the one the other one is **forked from**

call it "fork" or similar

git.go
```diff
@@ -0,0 +1,244 @@
...
+			return fmt.Errorf("couldn't add remote (%s), error: %v", config.PRRepositoryCloneURL, err)
+		}
+
+		if err := runWithRetry(func() *command.Model {
+			return Git.Fetch("upstream", config.Branch)
```
viktorbenei: same here

git.go
```diff
@@ -0,0 +1,244 @@
...
+		}); err != nil {
+			return fmt.Errorf("fetch Pull Request branch failed (%s), error: %v",
+				config.Branch, err)
+		}
+
+		if err := run(Git.Merge("upstream/" + config.Branch)); err != nil {
```
viktorbenei: an here of course - and everywhere where "upstream" is used, as we changed the order

git.go
```diff
@@ -0,0 +1,244 @@
...
+			return Git.Fetch("--depth=" + config.CloneDepth)
+		}
+		return Git.Fetch()
+	}); err != nil {
+		return fmt.Errorf("Fetch failed, error: %v", err)
+	}
+
+	if err := run(Git.Checkout(arg)); err != nil {
+		if config.CloneDepth == "" {
+			return fmt.Errorf("checkout failed (%s), error: %v", checkoutArg, err)
```
viktorbenei: Where does `checkoutArg` come from? this function's input is `arg` not `checkoutArg`, and the first Checkout does use `arg` and not `checkoutArg`

Please avoid global variable as much as possible, other than maybe a single `config`!!

main.go
```diff
@@ -1,98 +1,50 @@
...
-		PullRequestID:          os.Getenv("pull_request_id"),
-		BranchDest:             os.Getenv("branch_dest"),
-		PullRequestMergeBranch: os.Getenv("pull_request_merge_branch"),
-		ResetRepository:        os.Getenv("reset_repository"),
-		ManualMerge:            os.Getenv("manual_merge"),
+var (
+	config Config
+	// Git ...
+	Git         *git.Git
+	checkoutArg string
```
viktorbenei: please do not use global variables if not necessary!

main.go
```diff
@@ -1,98 +1,50 @@
...
-		PullRequestURI:         os.Getenv("pull_request_repository_url"),
-		PullRequestID:          os.Getenv("pull_request_id"),
-		BranchDest:             os.Getenv("branch_dest"),
-		PullRequestMergeBranch: os.Getenv("pull_request_merge_branch"),
-		ResetRepository:        os.Getenv("reset_repository"),
-		ManualMerge:            os.Getenv("manual_merge"),
+var (
+	config Config
+	// Git ...
+	Git         *git.Git
```
viktorbenei: I'd actually remove this one too, and possibly even config; should be passed to the functions if required as a param instead. Makes testing way easier as well.

main.go
```diff
@@ -1,98 +1,50 @@
...
-	log.Printf("- BuildAPIToken: %s", configs.BuildAPIToken)
-}
-
-func (configs ConfigsModel) validate() error {
-	if configs.CloneIntoDir == "" {
-		return errors.New("no CloneIntoDir parameter specified")
-	}
-	if configs.RepositoryURL == "" {
-		return errors.New("no RepositoryURL parameter specified")
+func printLog(format, env string) error {
```
viktorbenei: please rename this function, I thought this is just a simple print func but it also exports with envman. Not clear from its name.

main.go
```diff
@@ -102,308 +54,94 @@ func exportEnvironmentWithEnvman(keyStr, valueStr string) error {
...
-				os.Exit(1)
-			}
-			if err := fetchWithRetry(git); err != nil {
-				log.Errorf("Failed, error: %s", err)
-				os.Exit(1)
-			}
-		} else {
-			os.Exit(1)
-		}
+	if err := os.MkdirAll(config.CloneIntoDir, 0700); err != nil {
```
viktorbenei: please never use `700` unless required, this will make it only accessible for the user, which might conflict with other, usually system/daemon tools

viktorbenei: in general `755` should be the safe default ; if you do a `git clone` in the terminal, or just a `mkdir -p` then the created dir will be `755`

PR #56, opened 2017-12-14
Add deprecate warning for reset repository value options

kdobmayer:


config.go
```diff
@@ -75,11 +75,15 @@ func newConfig() (Config, []error) {
 	}
 
 	// bools
-	err = input.ValidateWithOptions(os.Getenv("reset_repository"), "yes", "no")
+	resetRepo := os.Getenv("reset_repository")
+	if resetRepo == "Yes" || resetRepo == "No" {
+		log.Warnf("\nInput values 'Yes' and 'No' for reset_repository are DEPRECATED in favor of 'yes' and 'no', these value options will be removed in Git Clone step version 4.1.0\n")
```
godrei: for reset_repository are -> for reset_repository input are

these value options will be removed in Git Clone step version 4.1.0 -> these value options will be removed in version 4.1.0

step.yml
```diff
@@ -68,6 +68,8 @@ inputs:
     opts:
       category: Debug
       summary: Reset repository contents with git reset --hard HEAD and git clean -f after repository updated
+      description: |-
+        Input values `Yes` and `No` are DEPRECATED, these value options will be removed in version 4.1.0
```
godrei: Input values `Yes` and `No` are DEPRECATED, these value options will be removed in version 4.1.0. Please use `yes` and `no` instead.

PR #61, opened 2017-12-14
Ignore depth if PR

kdobmayer:


git.go
```diff
@@ -137,11 +137,8 @@ func isPrivate(repoURL string) bool {
 	return strings.HasPrefix(repoURL, "git")
 }
 
-func autoMerge(gitCmd git.Git, mergeBranch, branchDest, buildURL, apiToken string, depth, id int) error {
+func autoMerge(gitCmd git.Git, mergeBranch, branchDest, buildURL, apiToken string, id int) error {
```
viktorbenei: Depth can be used in case of auto merge, it doesn't matter here,only in case of manual merge

viktorbenei:
Otherwise 👍😉

PR #64, opened 2017-12-19
Call git merge if the checkout target is a branch

kdobmayer:


main.go
```diff
@@ -92,6 +92,11 @@ func mainE() error {
 		if err := checkout(gitCmd, checkoutArg, config.CloneDepth); err != nil {
 			return fmt.Errorf("checkout (%s): %v", checkoutArg, err)
 		}
+		if config.Commit == "" && config.Branch != "" {
+			if err := run(gitCmd.Merge("origin/" + config.Branch)); err != nil {
```
viktorbenei: add a comment about what this is / why we do this here

git.go
```diff
@@ -183,6 +183,9 @@ func manualMerge(gitCmd git.Git, repoURL, prRepoURL, branch, commit, branchDest
 	if err := run(gitCmd.Checkout(branchDest)); err != nil {
 		return fmt.Errorf("checkout failed (%s), error: %v", branchDest, err)
 	}
+	if err := run(gitCmd.Merge("origin/" + branchDest)); err != nil {
```
viktorbenei: add a comment about what this is / why we do this here

main.go
```diff
@@ -92,6 +92,12 @@ func mainE() error {
 		if err := checkout(gitCmd, checkoutArg, config.CloneDepth); err != nil {
 			return fmt.Errorf("checkout (%s): %v", checkoutArg, err)
 		}
+		// Update branch: 'git fetch' followed by a 'git merge' is the same as 'git pull'.
+		if config.Commit == "" && config.Branch != "" {
+			if err := run(gitCmd.Merge("origin/" + config.Branch)); err != nil {
+				return fmt.Errorf("merge %q: %v", checkoutArg, err)
```
viktorbenei: `checkoutArg` used in the error but in the actual command `config.Branch` is used; should be unified

PR #68, opened 2018-01-16
Use pull/x/head branch then merge instead of pull/x/merge

kdobmayer:


git.go
```diff
@@ -155,18 +155,25 @@ func autoMerge(gitCmd git.Git, mergeBranch, branchDest, buildURL, apiToken strin
 	}
 
 	if mergeBranch != "" {
+		arg := strings.TrimSuffix(mergeBranch, "/merge")
```
viktorbenei: can you please write a test for these? it starts to get complicated, quite a bit

I'd suggest you to move this if case / logic into a function which returns the fetch, checkout etc. params but does not perform them - that way we can write unit tests for these

PR #69, opened 2018-01-24
Set inputs to read-only

kdobmayer:


yM1ndst0rM:
Could you provide some context for this change?

Im trying to checkout two different git repos into two different working directories within one Bitrise build and found out that this relatively recent change prevents me from using this step several times with slightly different `Git repository URL` and `Clone destination (local) directory path`.

I did not find information about that in the Bitrise support community and came here by looking through the change logs for the build step.

Please point me to a discussion or reasoning behind this change if a public one exists.
What should be the proper way to implement this build process provided that I'm not allowed to change the input variables?

viktorbenei:
@yM1ndst0rM sure! In short, please do not use this step to clone any other repository other than the primary one of the workflow.

Why? Because this step is configured by default to use configurations passed to the build, like branch, commit, pull request infos etc.

So if you just set it to another repository and then you start a pull request build for the primary repo (the one you connected on bitrise.io) then you'll end up git cloning the secondary repo with similar params e.g. the pull request branch, which is not what you want.

For any other repo other than the primary please use a Script step and do `git clone` there, the way you intend to; or add the secondary repo as a submodule to the primary so that a single git clone step can clone / checkout both.

If you'd have any questions just let us know! :)

PR #96, opened 2019-07-05
TOOL-57 #comment Updated to trim commit message body to 72 if longer.

richard-bogdan-bitrise:


main.go
```diff
@@ -31,19 +31,34 @@ type config struct {
...
+
 	log.Printf("=> %s\n   value: %s\n", env, l)
 	if err := exportEnvironmentWithEnvman(env, l); err != nil {
 		return fmt.Errorf("envman export, error: %v", err)
 	}
 	return nil
 }
 
+func checkForTrim(value string) string {
+	const maxLength = 72
```
BirmacherAkos: Move the `const maxLength` to global. Usualy we don't like to use global variables, but in later it will be much easier to find and modify the value.

For example:

```
import (
    "fmt"
    ...
)

const 
(
    maxCommitBodyLenght = 72
)

main() {
    ...
    ...
```
}

main.go
```diff
@@ -31,19 +31,34 @@ type config struct {
...
+	}
+
 	log.Printf("=> %s\n   value: %s\n", env, l)
 	if err := exportEnvironmentWithEnvman(env, l); err != nil {
 		return fmt.Errorf("envman export, error: %v", err)
 	}
 	return nil
 }
 
+func checkForTrim(value string) string {
```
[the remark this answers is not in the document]
BirmacherAkos: And in this case, I think you don't need to write a specific method for it., Instead use it locally:

```
if trim && len(value) > maxLength {
        trimmedValue := l[: maxLength]
	log.Printf("Value %s\n trimmed to =>\n%s", l, trimmedValue)
	l =  trimmedValue
}
```

main.go
```diff
@@ -132,7 +147,12 @@ func mainE() error {
 			`%cn`: "GIT_CLONE_COMMIT_COMMITER_NAME",
 			`%ce`: "GIT_CLONE_COMMIT_COMMITER_EMAIL",
 		} {
-			if err := printLogAndExportEnv(gitCmd, format, env); err != nil {
+			if env == "GIT_CLONE_COMMIT_MESSAGE_BODY" {
```
BirmacherAkos: In Go you can evaluate the bool expression in the function's method:
Instead of:
```
if env == "GIT_CLONE_COMMIT_MESSAGE_BODY" {
	if err := printLogAndExportEnv(gitCmd, format, env, true); err != nil {
		return fmt.Errorf("gitCmd log failed, error: %v", err)
	}
}

if err := printLogAndExportEnv(gitCmd, format, env, false); err != nil {
	return fmt.Errorf("gitCmd log failed, error: %v", err)
}
```

Use:
```

if err := printLogAndExportEnv(gitCmd, format, env,(env == "GIT_CLONE_COMMIT_MESSAGE_BODY")); err != nil {
	return fmt.Errorf("gitCmd log failed, error: %v", err)
}
```

richard-bogdan-bitrise: Fixed.

main.go
richard-bogdan-bitrise: Fixed.

main.go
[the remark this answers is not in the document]
richard-bogdan-bitrise: Thanks for the explanation. Fixed.

richard-bogdan-bitrise:
Added a test case too. Please check new version. Added PR:
https://github.com/bitrise-io/git-clone-test/pull/9
This is needed for the test. If you have no additional comments, please merge that, and I have to update here in the new test the commit hash and branch before merging.

main.go
```diff
@@ -31,12 +31,22 @@ type config struct {
 	ManualMerge      bool   `env:"manual_merge,opt[yes,no]"`
 }
 
-func printLogAndExportEnv(gitCmd git.Git, format, env string) error {
+const (
+	maxCommitBodyLenght = 72
```
usmonster: typo; should be `maxCommitBodyLength`

main.go
```diff
@@ -31,12 +31,22 @@ type config struct {
...
+)
+
+func printLogAndExportEnv(gitCmd git.Git, format, env string, trim bool) error {
 	l, err := output(gitCmd.Log(format))
 	if err != nil {
 		return err
 	}
 
+	if trim && len(l) > maxCommitBodyLenght {
+		trimmedValue := l[:maxCommitBodyLenght]
```
usmonster: Would've liked to append an indicator that it was trimmed, like `…`

main.go
```diff
@@ -132,7 +142,7 @@ func mainE() error {
 			`%cn`: "GIT_CLONE_COMMIT_COMMITER_NAME",
 			`%ce`: "GIT_CLONE_COMMIT_COMMITER_EMAIL",
 		} {
-			if err := printLogAndExportEnv(gitCmd, format, env); err != nil {
+			if err := printLogAndExportEnv(gitCmd, format, env, (env == "GIT_CLONE_COMMIT_MESSAGE_SUBJECT" || env == "GIT_CLONE_COMMIT_MESSAGE_BODY")); err != nil {
```
usmonster: What's the motivation behind this change, actually? And how can I override this behavior?

(The real "question behind the question" is: why are my Slack notifications now showing truncated commit messages, and how can I get the full ones back? Please let me know if I should open a separate issue, or ask on the forum.)

Thanks!

main.go
richard-bogdan-bitrise: Fixed.

main.go
richard-bogdan-bitrise: Great idea, added indicator.

main.go
richard-bogdan-bitrise: This has been updated, now it is limited by the maximum item size in envman. From one side it is much more than the 72 chars, on the other side, you can configure it by increasing envman item max size.
The motivation behind was that, some commit message or subject was extremely long, and the build failed, as envman was unable to save them, because they were longer than the limit.

usmonster: Thanks for the explanation and the update!

PR #98, opened 2019-07-26
TOOL-57 Updated max commit message length to reflect to envman max size.

richard-bogdan-bitrise:
PR Dependency: https://github.com/bitrise-io/git-clone-test/pull/10


main.go
```diff
@@ -32,18 +33,20 @@ type config struct {
 }
 
 const (
-	maxCommitBodyLenght = 72
+	trimEnding = "..."
 )
 
-func printLogAndExportEnv(gitCmd git.Git, format, env string, trim bool) error {
+var maxCommitMessageLength int
```
BirmacherAkos: Do not use global variable if you can.
The previous `maxCommitBodyLenght` was global because it had a default value, and it was easier to modify it.
But because the `maxCommitMessageLength` is a variable with no default value I would recommend to use a local variable instead and pass it between the functions.

richard-bogdan-bitrise: If I do not use a global variable, then I have to pass it to printLogAndExportEnv every time that method is called, but it is only needed in only 2 cases.

Other way is to make it a getter, and call it when needed, but it would be still redundant.

BirmacherAkos:
The CI is failing, because the `environment list size` is larger than the max allowed size.
Screenshot 2019-08-05 at 12 13 14

BirmacherAkos:
@godrei 
Please check/merge the PR.

main.go
godrei: i agree with @BirmacherAkos :
- modify `initCommitMessageLength`: it should just read the envman config and return the maxLenght
- printLogAndExportEnv is only called in the mainE, where you call the current init initCommitMessageLength

main.go
```diff
@@ -60,14 +63,27 @@ func exportEnvironmentWithEnvman(keyStr, valueStr string) error {
 	return cmd.Run()
 }
 
+func initCommitMessageLength() error {
+	configs, err := envman.GetConfigs()
+	if err != nil {
+		return err
+	}
+
+	maxCommitMessageLength = configs.EnvBytesLimitInKB*1024 - len(trimEnding)
```
godrei: NOTE: it would be better if `maxCommitMessageLength` would be the max commit message length regardless of the used separator.

main.go
```diff
@@ -32,18 +33,20 @@ type config struct {
...
 	if err != nil {
 		return err
 	}
 
-	if trim && len(l) > maxCommitBodyLenght {
-		trimmedValue := l[:maxCommitBodyLenght]
-		log.Printf("Value %s\n trimmed to =>\n%s", l, trimmedValue)
+	if trimmableVariable && len(l) > maxCommitMessageLength {
+		trimmedValue := l[:maxCommitMessageLength] + trimEnding
+		log.Printf("Value %s trimmed", env)
```
godrei: pls explain why we trim the value, without that this log does not really make sense

main.go
```diff
@@ -32,18 +33,20 @@ type config struct {
...
 const (
-	maxCommitBodyLenght = 72
+	trimEnding = "..."
 )
 
-func printLogAndExportEnv(gitCmd git.Git, format, env string, trim bool) error {
+var maxCommitMessageLength int
+
+func printLogAndExportEnv(gitCmd git.Git, format, env string, trimmableVariable bool) error {
 	l, err := output(gitCmd.Log(format))
```
godrei: pls fix the variable names if you are already touching this function: https://github.com/golang/go/wiki/CodeReviewComments#variable-names

main.go
richard-bogdan-bitrise: Added details.

main.go
richard-bogdan-bitrise: Updated code.

main.go
richard-bogdan-bitrise: renamed local variables

main.go
```diff
@@ -32,19 +33,21 @@ type config struct {
 }
 
 const (
-	maxCommitBodyLenght = 72
+	trimEnding = "..."
 )
 
-func printLogAndExportEnv(gitCmd git.Git, format, env string, trim bool) error {
+var maxCommitMessageLength int
```
godrei: i agree with @BirmacherAkos :
- modify `initCommitMessageLength`: it should just read the envman config and return the maxLenght
- printLogAndExportEnv is only called in the mainE, where you call the current init initCommitMessageLength, so you can easily pass the max message lenght as local variable
- instead of `trimmableVariable bool` input, accept the max env length

PR #114, opened 2020-09-04
Add option to disable merge origin/master

DamienBitrise:
Not merging to master before building can significantly speed up the git clone. ie. git merge origin/master

image

All settings default to current functionality

This means when the step is updated nothing will change unless users changes this setting

### DIFF
--- a/bitrise.yml
+++ b/bitrise.yml
@@ -43,6 +43,7 @@ workflows:
     - _test_commit_logs
     - _test_hosted_git_notfork
     - _test_unrelated_histories
+    - _test_unrelated_histories_with_options
     - _test_hosted_git_ssh_prefix
 
   _test_submodule:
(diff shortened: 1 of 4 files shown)


Szabadember (approved):

git.go
```diff
@@ -230,12 +230,18 @@ func autoMerge(gitCmd git.Git, mergeBranch, branchDest, buildURL, apiToken strin
...
 		return fmt.Errorf("fetch failed, error: %v", err)
 	}
-	if err := pull(gitCmd, branchDest); err != nil {
-		return fmt.Errorf("pull failed (%s), error: %v", branchDest, err)
+	if autoMerge {
+		if err := pull(gitCmd, branchDest); err != nil {
+			return fmt.Errorf("pull failed (%s), error: %v", branchDest, err)
+		}
+	} else {
+		if err := run(gitCmd.Checkout(branchDest)); err != nil {
```
Szabadember, line 242: branchDest = master; is this the correct branch?

Szabadember: Maybe $Branch would be the correct branch?

step.yml
```diff
@@ -124,6 +124,16 @@ inputs:
       value_options:
       - "yes"
       - "no"
+  - auto_merge: "yes"
+    opts:
+      category: Debug
+      title: Merge origin/master
+      summary:  Merge origin/master 
+      description: |-
+        Merge origin/master always
```
Szabadember, line 133: Master is not always the PR target, description should say that this flag merges our source branch into the destination branch.

step.yml
```diff
@@ -124,6 +124,16 @@ inputs:
       value_options:
       - "yes"
       - "no"
+  - auto_merge: "yes"
```
Szabadember, line 127: This name seems a bit confusing to me because something is either manual or automatic but not both. We need a better name for this flag. What about `checkout_merged_state`?

step.yml
```diff
@@ -124,6 +124,16 @@ inputs:
...
+    opts:
+      category: Debug
+      title: Merge origin/master
```
Szabadember, line 130: Checkout merged state

step.yml
```diff
@@ -124,6 +124,16 @@ inputs:
...
+      summary:  Merge origin/master
```
Szabadember, line 131: Flag indicating if merged state should be checked out instead of just the source branch.

step.yml
Szabadember: Switching this flag to `true` indicates that the source branch should be merged into the destination branch. This is useful if you want to build/test the `integrated state` (e. g. feature branch merged into master) instead of just the source branch.

Szabadember (changes requested):

git.go
damo1884: branchDest is the existing branch, no changes here just that it does not call 

`run(gitCmd.Merge("origin/" + branchDest))`

DamienBitrise: branchDest is the existing branch, no changes here just that it does not call 

`run(gitCmd.Merge("origin/" + branchDest))`

step.yml
DamienBitrise: Yep I wasn't too keep on the name either

bitrise.yml
```diff
@@ -544,6 +545,40 @@ workflows:
...
+        - tag: ""
+        - branch: ""
+        - clone_depth: "50"
+        - manual_merge: "no"
+        - checkout_merged_state: "no"
+    - script:
+        inputs:
+        - content: |-
+            #!/bin/bash
+            echo "GIT_CLONE_COMMIT_HASH: ${GIT_CLONE_COMMIT_HASH}"
```
Szabadember, line 570: Maybe we could do a commit hash comparison as part of this test?

git.go
Szabadember: I think that the `autoMerge == true` case is correct but if autoMerge is disabled (else clause) the step will check out master (branchDest) instead of the feature banch (branch).

The PR is opened for merging `branch` -> `branchDest`, e.g.: `feature1` -> `master`.
`autoMerge == true` should check out `feature1 merge into master` and `autoMerge == false` should check out `feature1`. Is that correct @DamienBitrise  @godrei ?

Szabadember (changes requested):

git.go
DamienBitrise: Pull() does the following

`if err := run(gitCmd.Checkout(branchDest)); err != nil {
		return fmt.Errorf("checkout failed (%s), error: %v", branchDest, err)
	}
	return run(gitCmd.Merge("origin/" + branchDest))`

So this is all that is different between the if else

`run(gitCmd.Merge("origin/" + branchDest))`

Szabadember: So as @DamienBitrise pointed out, branchDest  is the correct one, see: https://bitrise.atlassian.net/wiki/spaces/EnterpriseSales/pages/700514342/Git+Clone+-+Performance+Analysis

At this point, we either check out master or check out master + merge master but what is the value of merging master into master? @godrei @trapacska

DamienBitrise:
Abandoning

PR #115, opened 2020-09-04
Add option to not check out tags

DamienBitrise:
**NOTE:** This must not be merged before https://github.com/bitrise-steplib/steps-git-clone/pull/114 and will require merge conflicts to be resolved once 114 is merged

Added the ability to optionally disable checking out tags

image

All settings default to current functionality

This means when the step is updated nothing will change unless users changed this setting

### DIFF
--- a/step.yml
+++ b/step.yml
@@ -124,6 +124,16 @@ inputs:
       value_options:
       - "yes"
       - "no"
+  - fetch_tags: "yes"
+    opts:
+      category: Debug
+      title: Fetch tags
+      summary: Adds --no-tags to fetch commands
+      description: |-
+        Adds --no-tags to fetch commands
+      value_options:
+      - "yes"
+      - "no"
   - build_url: "$BITRISE_BUILD_URL"
     opts:
       category: Debug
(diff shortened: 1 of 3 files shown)


step.yml
```diff
@@ -124,6 +124,16 @@ inputs:
       value_options:
       - "yes"
       - "no"
+  - fetch_tags: "yes"
+    opts:
+      category: Debug
+      title: Fetch tags
+      summary: Adds --no-tags to fetch commands
```
Szabadember, line 131: This could be confusing because setting the flag excludes `--no-tags` instead of adding it. Maybe we could call this `no_tags` instead or clarify the summary a little bit more? `Disabling this flag adds --no-tags to fetch commands`
@godrei which solution should we go with?

Szabadember (changes requested):

step.yml
DamienBitrise: I agree it should be the reverse and have the setting disabled by default

DamienBitrise:
Abandoning

PR #123, opened 2020-10-19
Get list of branches

doomsayer13:


gitclone/git.go
```diff
@@ -280,6 +292,19 @@ func checkout(gitCmd git.Git, arg, branch string, depth int, isTag bool) *step.E
 		}
 		return gitCmd.Fetch(opts...)
 	}); err != nil {
+		if branch != "" {
+			branches, branchesErr := listBranches(gitCmd)
+			if branchesErr == nil {
```
godrei, line 297: We add this logic, to let the user know if wrong branch was provided.
- This means we should check here if the given branch is contained in the `branches` slice, if yes we should not send the new error (`newStepErrorWithRecommendations `)
- We could include this information in the error we return (and the step prints this also), so please update `shortMsg` of the error: something like this: `fmt.Errorf("fetch failed: invalid branch selected: %s, available branches: %s: %v", branch, strings.Join(branches, ", "), err)`, so that both the step will print more informative error message

godrei (changes requested):
Pls check my comment

gitclone/git.go
doomsayer13: This means that I have to write a `Contains` or `Find` function? What should we do the `branch` is in the `branches` list?

doomsayer13: Done ;)

gitclone/git.go
```diff
@@ -266,6 +266,28 @@ func manualMerge(gitCmd git.Git, repoURL, prRepoURL, branch, commit, branchDest
...
+	branches, err := gitCmd.Branch("-r").RunAndReturnTrimmedCombinedOutput()
+	if err != nil {
+		return nil, err
+	}
+
+	return strings.Split(branches, " "), nil
+}
+
+// Contains tells whether slice contains val.
+func Contains(slice []string, val string) bool {
```
lpusok, line 282: There is  a helper function for this in sliceutil pacakge: https://github.com/bitrise-io/go-utils/blob/6cc2aa4d248adef509996ac484bffd38856ddf80/sliceutil/sliceutil.go#L32

gitclone/git.go
```diff
@@ -280,6 +302,22 @@ func checkout(gitCmd git.Git, arg, branch string, depth int, isTag bool) *step.E
 		}
 		return gitCmd.Fetch(opts...)
 	}); err != nil {
+		if branch != "" {
+			branches, branchesErr := listBranches(gitCmd)
+			if branchesErr == nil {
+				contains := Contains(branches, branch)
+				if !contains {
```
lpusok, line 309:
```suggestion
			if branchesErr == nil && sliceutil.IsStringInSlice(branches, branch) {
```

godrei (approved):

PR #124, opened 2020-10-20
Including git command output in error message.

lpusok:
Include git command output in errors, instead of the exit code.

### DIFF
--- a/bitrise.yml
+++ b/bitrise.yml
@@ -29,6 +29,7 @@ workflows:
     - errcheck:
     - go-test:
     after_run:
+    - _test_error
     - _test_submodule
     - _test_no_checkout
     - _test_checkout_commit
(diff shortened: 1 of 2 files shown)


bitrise.yml
```diff
@@ -78,6 +79,53 @@ workflows:
         inputs:
         - dir_to_check: $CLONE_INTO_DIR
 
+  error_test:
```
godrei, line 82: I do not see the added value of this test, what we changed here is commands are returning the output in case of exist status error

gitclone/git.go
```diff
@@ -107,7 +111,18 @@ func getDiffFile(buildURL, apiToken string, prID int) (string, error) {
...
-	return c.SetStdout(os.Stdout).SetStderr(os.Stderr).Run()
+	var buffer bytes.Buffer
+	outWriter := io.MultiWriter(os.Stdout, &buffer)
+
+	err := c.SetStdout(outWriter).SetStderr(outWriter).Run()
+	if err != nil {
+		if errorutil.IsExitStatusError(err) {
+			return errors.New(buffer.String())
+		}
+		return fmt.Errorf("command execution failed, %s", err)
```
godrei, line 122: i do not see the added value of `command execution failed` prefix, previously we just returned and logged the error

gitclone/git.go
```diff
@@ -107,7 +111,18 @@ func getDiffFile(buildURL, apiToken string, prID int) (string, error) {
...
+	}
+
+	return nil
 }
 
 func output(c *command.Model) (string, error) {
```
godrei, line 128: this function should be also updated as in case of an error we just print the error

godrei: also `listBranches` function could also be updated

godrei (changes requested):
pls check my notes

gitclone/git.go
lpusok: True, this only happens if the error is not exit status error tough.

godrei (approved):

PR #126, opened 2020-11-03
Scanner error mapping [ACT-630][ACT-865]

zoltanszabo-bitrise:
Jira ticket: https://bitrise.atlassian.net/browse/ACT-630
Please see for error details: https://bitrise.atlassian.net/wiki/spaces/ACT/pages/833782484/Scanner+error+messages


gitclone/steperror.go
```diff
@@ -1,11 +1,178 @@
...
+	for pattern, handler := range m.handlers {
+		re := regexp.MustCompile(pattern)
+		if re.MatchString(msg) {
+			matches := re.FindStringSubmatch((msg))
+
+			if len(matches) > 1 {
+				matches = matches[1:]
+			}
+
+			if matches != nil {
```
godrei, line 39: i think if `re.MatchString(msg)` is true, matches can not be nil

gitclone/steperror.go
```diff
@@ -1,11 +1,178 @@
...
+}
+
+// Run ...
+func (m *PatternErrorMatcher) Run(msg string) step.Recommendation {
+	for pattern, handler := range m.handlers {
+		re := regexp.MustCompile(pattern)
+		if re.MatchString(msg) {
+			matches := re.FindStringSubmatch((msg))
+
+			if len(matches) > 1 {
```
godrei, line 35: i suggest introducing a new variable to store the match groups, this way we express that the `handler` expects the match groups (and not the first item of `matches`: the whole string matches to the pattern) as input.

by introducing the match groups variable, we can express here, that some handler receives nil/empty slice

gitclone/steperror.go
```diff
@@ -1,11 +1,178 @@
...
+// DetailBuilder ...
+type DetailBuilder = func(...string) Detail
+
+// PatternErrorMatcher ...
+type PatternErrorMatcher struct {
+	defaultHandler DetailBuilder
+	handlers       map[string]DetailBuilder
+}
+
+func newPatternErrorMatcher(defaultHandler DetailBuilder, handlers map[string]DetailBuilder) *PatternErrorMatcher {
```
godrei, line 19: calling a DetailBuilder as `handler` is a bit confusing, why not `defaultBuilder` and `patternToBuilder`?

gitclone/steperror.go
```diff
@@ -1,11 +1,178 @@
...
+func newDetailedErrorRecommendation(detail Detail) step.Recommendation {
+	return step.Recommendation{
+		"DetailedError": map[string]string{
+			"Title":       detail.Title,
+			"Description": detail.Description,
+		},
+	}
+}
+
+func checkoutFailedGenericDetail(params ...string) Detail {
```
godrei, line 117: these are constructors of `Detail`, i suggest using `new` prefix and `Detail` suffix for these contractors.

godrei: Also by convention, first please define the struct, then the constructors, then struct functions.

gitclone/steperror.go
```diff
@@ -1,11 +1,178 @@
...
+	return Detail{
+		Title:       fmt.Sprintf("We couldn’t connect to '%s'.", host),
+		Description: "Please abort the process, double-check your repository URL and try again.",
+	}
+}
+
+func fetchFailedSamlSSOError(params ...string) Detail {
+	return Detail{
+		Title:       "To access this repository, you need to use SAML SSO.",
+		Description: `Please abort the process, update your SSH settings and try again. You can find out more about <a target="_blank" href="https://docs.github.com/en/free-pro-team@latest/github/authenticating-to-github/authorizing-an-ssh-key-for-use-with-saml-single-sign-on">using SAML SSO in the Github docs</a>.`,
```
godrei, line 166: is our frontend prepared to present HTML content properly?

gitclone/steperror.go
```diff
@@ -1,11 +1,178 @@
 package gitclone
 
-import "github.com/bitrise-io/bitrise-init/step"
+import (
```
godrei, line 3: I suggest splitting this file into two: 
- error_matcher.go (includes `PatternErrorMatcher ` and related types `DetailBuilder `, `Detail `, ..)
- steperror.go (includes `mapRecommendation`, step specific implementations)

so that later we can easily move the shared code to a new package.

godrei (changes requested):
Nice solution! I left some comments, please check.
Also please check the CI issues.

gitclone/steperror.go
```diff
@@ -1,11 +1,178 @@
...
+			// `ssh: connect to host (.+) port \d+: Connection refused`
+			// `ssh: connect to host (.+) port \d+: Network is unreachable`
+			`ssh: connect to host (.+) port \d+:`:                                fetchFailedCouldConnectError,
+			`ssh: Could not resolve hostname (.+): Name or service not known`:    fetchFailedCouldConnectError,
+			`fatal: unable to access '.+': Could not resolve host: (\S+)`:        fetchFailedCouldConnectError,
+			`ERROR: The \x60(.+)' organization has enabled or enforced SAML SSO`: fetchFailedSamlSSOError,
+		})
+}
+
+func mapRecommendation(tag string, err error) step.Recommendation {
```
godrei, line 84: as we only need the error message, I suggest passing a string only as input (this will help with the CI issue too)

gitclone/steperror.go
zoltanszabo-bitrise: Probably you are right.

gitclone/steperror.go
zoltanszabo-bitrise: Yep, good idea.
Since we are having at least [search_string], and we left off the first param, we will end up with an empty slice if there is no match group defined in the regex.
`matches[1:]` returns either `[]` or [`match1, match2, ...]`

gitclone/steperror.go
zoltanszabo-bitrise: Not sure, but we will have to change it anyway to support the new DetailedError structure, so we can make sure to handle it.

gitclone/steperror.go
```diff
@@ -1,11 +1,178 @@
...
+	}
+
+	detail := m.defaultHandler(msg)
+	return newDetailedErrorRecommendation(detail)
+}
+
+func newCheckoutFailedPatternErrorMatcher() *PatternErrorMatcher {
+	return newPatternErrorMatcher(
+		checkoutFailedGenericDetail,
+		map[string]DetailBuilder{},
```
godrei, line 53: no need for initializing an empty map, pass nil instead.

errormapper/errormapper.go
```diff
@@ -0,0 +1,68 @@
+package errormapper
```
mateherber, line 1: Shouldn't this be in an other repo if it is used from different steps? I think we shouldn't introduce dependency between steps just to use "utility" kind of stuff

chreno (approved):

godrei (approved):

PR #132, opened 2020-12-07
Fix: Do not add branch to fetch when trying to checkout commit

mateherber:
We wanted to improve on our git clone times by always fetching the
provided branch. This approach works with tags but it doesn't work
always with commits.

* Added test to exploit this case

[STEP-25]

[STEP-25]: https://bitrise.atlassian.net/browse/STEP-25

### DIFF
--- a/bitrise.yml
+++ b/bitrise.yml
@@ -33,6 +33,7 @@ workflows:
     - _test_submodule
     - _test_no_checkout
     - _test_checkout_commit
+    - _test_checkout_commit_with_other_branch
     - _test_checkout_tag
     - _test_checkout_branch
     - _test_checkout_pull_request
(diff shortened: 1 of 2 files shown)


bitrise.yml
```diff
@@ -190,6 +191,38 @@ workflows:
...
+            - branch: "master"
+            - pull_request_id: ""
+            - pull_request_merge_branch: ""
+            - pull_request_repository_url: ""
+            - branch_dest: ""
+            - clone_depth: ""
+      - script:
+          inputs:
+            - content: |-
+                #!/bin/bash
```
adborbas, line 213: This script step is repeated multiple times. Could we either extract it for reuse now or add it to our list of improvements for the git clone step?

gitclone/git.go
```diff
@@ -363,7 +363,7 @@ func checkout(gitCmd git.Git, arg, branch string, depth int, isTag bool) *step.E
 		if isTag {
 			opts = append(opts, "--tags")
 		}
-		if branch != "" {
+		if branch == arg || (branch != "" && isTag) {
```
lpusok, line 366: Can you add a "test" case for when both branch and tag is specified? Why does it differ from the commit one?

mateherber: Yes

lpusok (approved):
Looks correct, approved. There is still room to improve the tag checkout in the future, to only fetch the needed tag (similar to how is done with branches, see https://stackoverflow.com/a/45338695)

adborbas (approved):

PR #133, opened 2020-12-08
Clone step will not try to merge if there is commit or tag given

mateherber:
If commit or tag is given we check out them so we don't need to merge to
a branch with similar name

* Added test case

[STEP-36]

[STEP-36]: https://bitrise.atlassian.net/browse/STEP-36

### DIFF
--- a/bitrise.yml
+++ b/bitrise.yml
@@ -47,6 +47,7 @@ workflows:
     - _test_checkout_commit_with_other_branch
     - _test_checkout_tag
     - _test_checkout_tag_with_other_branch
+    - _test_checkout_tag_with_equally_named_branch
     - _test_checkout_branch
     - _test_checkout_pull_request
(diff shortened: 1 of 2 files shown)


adborbas (dismissed):

lpusok (changes requested):
There is branch == arg condition inside checkout(). When both tag and branch provided, it checks out the branch, not the tag. is it intended?

mateherber:
> There is branch == arg condition inside checkout(). When both tag and branch provided, it checks out the branch, not the tag. is it intended?

It involves the fetch only when both branch and tag will be fetched. The checkout will be `git checkout tag_branch_name` which will checkout the branch if it is present

lpusok (commented):
discussed, lgtm

lpusok (dismissed):

godrei (approved):

PR #134, opened 2021-02-24
Introduce unit test for checkout.

godrei:
### Checklist

### Context

In this PR I plan to introduce unit testing for the checkout functionality of the step.

### Changes

- Introduce `CmdRunner` interface for `Output`, `Run`, and `RunWithRetry` functions and move the current implementation to `DefaultRunner`, this allows us to mock the runner in unit tests.
- Introduce `MockRunner` which does not execute commands, but stores their PrintableCommandArgs()
- Simplify `runWithRetry`
- Move checkout functionality into `checkoutState`, so we can set up isolated unit tests

### DIFF
--- a/gitclone/git.go
+++ b/gitclone/git.go
@@ -28,6 +28,65 @@ const (
 	fetchFailedTag    = "fetch_failed"
 )
 
+var runner CmdRunner = DefaultRunner{}
+
+// CmdRunner ...
+type CmdRunner interface {
+	Output(c *command.Model) (string, error)
+	Run(c *command.Model) error
+	RunWithRetry(c *command.Model) error
+}
+
+// DefaultRunner ...
+type DefaultRunner struct {
+}
+
+// Output ...
+func (r DefaultRunner) Output(c *command.Model) (string, error) {
+	log.Infof("$ %s &> out", c.PrintableCommandArgs())
+
+	out, err := c.RunAndReturnTrimmedCombinedOutput()
+	if err != nil && errorutil.IsExitStatusError(err) {
+		return out, errors.New(out)
+	}
+
+	return out, err
+}
+
+// Run ...
+func (r DefaultRunner) Run(c *command.Model) error {
+	fmt.Println()
+	log.Infof("$ %s", c.PrintableCommandArgs())
+	var buffer bytes.Buffer
+
+	err := c.SetStdout(os.Stdout).SetStderr(io.MultiWriter(os.Stderr, &buffer)).Run()
+	if err != nil {
+		if errorutil.IsExitStatusError(err) {
+			return errors.New(strings.TrimSpace(buffer.String()))
+		}
+		return err
+	}
+
+	return nil
+}
+
+// RunWithRetry ...
+func (r DefaultRunner) RunWithRetry(c *command.Model) error {
+	return retry.Times(2).Wait(5).Try(func(attempt uint) error {
+		if attempt > 0 {
+			log.Warnf("Retrying...")
+		}
+
+		err := run(c)
+		if err != nil {
+			log.Warnf("Attempt %d failed:", attempt+1)
+			fmt.Println(err.Error())
+		}
+
+		return err
+	})
+}
+
 func isOriginPresent(gitCmd git.Git, dir, repoURL string) (bool, error) {
 	absDir, err := pathutil.AbsPath(dir)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


gitclone/git.go
```diff
@@ -115,46 +174,15 @@ func getDiffFile(buildURL, apiToken string, prID int) (string, error) {
...
-	err := c.SetStdout(os.Stdout).SetStderr(io.MultiWriter(os.Stderr, &buffer)).Run()
-	if err != nil {
-		if errorutil.IsExitStatusError(err) {
-			return errors.New(strings.TrimSpace(buffer.String()))
-		}
-		return err
-	}
-
-	return nil
+	return runner.Run(c)
```
adborbas, line 177: What's the point of extracting this single line to a function?

adborbas: Similar thing in line 181 and 185.

gitclone/gitclone_test.go
```diff
@@ -0,0 +1,411 @@
...
+
+const always = 9999
+const rawCmdError = "dummy_cmd_error"
+
+func Test_checkoutState(t *testing.T) {
+	tests := []struct {
+		name       string
+		cfg        Config
+		cmdOutputs map[string]commandOutput
+		want       *step.Error
```
adborbas, line 22: `wantErr` would more reflect that this is the expected error.

gitclone/gitclone_test.go
```diff
@@ -0,0 +1,411 @@
...
+		wantCmds   []string
+	}{
+		{
+			name: "Checkout commit",
+			cfg: Config{
+				RepositoryURL: "https://github.com/bitrise-io/steps-git-clone.git",
```
adborbas, line 28: Could you move the url and the commit to a constant?

gitclone/gitclone_test.go
```diff
@@ -0,0 +1,411 @@
...
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			mockRunner := newMockRunner(tt.cmdOutputs)
+			runner = mockRunner
+			got := checkoutState(git.Git{}, tt.cfg)
+			if !reflect.DeepEqual(got, tt.want) {
+				t.Errorf("checkoutState().err = (%#v), want %#v", got, tt.want)
+			}
+			// require.Equal(t, tt.want, got)
```
adborbas, line 344: Left a comment here.

gitclone/gitclone_test.go
```diff
@@ -0,0 +1,411 @@
...
+			},
+		},
+		{
+			name: "Checkout nonexistent branch",
+			cfg: Config{
+				RepositoryURL: "https://github.com/bitrise-io/steps-git-clone.git",
+				Branch:        "fake",
+			},
+			cmdOutputs: map[string]commandOutput{
+				`git "fetch" "origin" "refs/heads/fake"`: {failForCalls: always},
```
adborbas, line 87: You could also use a [mocking package](https://github.com/stretchr/testify/#mock-package) to stub the mock runner to return various outputs for various inputs. Example for setting up this particular mocking package: https://github.com/bitrise-steplib/steps-generic-file-storage/blob/90872ec8e32248f5bbc4ce695e86c4db398ff269/bulkfiledownloader/mockFileDownloader_test.go#L25

gitclone/gitclone_test.go
```diff
@@ -0,0 +1,411 @@
...
+
+	"github.com/bitrise-io/bitrise-init/step"
+	"github.com/bitrise-io/go-utils/command"
+	"github.com/bitrise-io/go-utils/command/git"
+)
+
+const always = 9999
+const rawCmdError = "dummy_cmd_error"
+
+func Test_checkoutState(t *testing.T) {
```
adborbas, line 17: Although I see the benefit of using table driven test for this scenario this test is more than 300 lines which makes it hard to understand. Could we group the scenarios in smaller functions?

gitclone/gitclone_test.go
```diff
@@ -0,0 +1,411 @@
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			mockRunner := newMockRunner(tt.cmdOutputs)
+			runner = mockRunner
+			got := checkoutState(git.Git{}, tt.cfg)
+			if !reflect.DeepEqual(got, tt.want) {
+				t.Errorf("checkoutState().err = (%#v), want %#v", got, tt.want)
+			}
+			// require.Equal(t, tt.want, got)
+			if !reflect.DeepEqual(mockRunner.Cmds(), tt.wantCmds) {
```
adborbas, line 345: I'm not sure what `reflect.DeepEqual` does but if just want to assert that two slices contains the same elements in the same order we could use the [testify/assert package](https://github.com/stretchr/testify/issues/759).

adborbas (changes requested):

gitclone/git.go
```diff
@@ -28,6 +28,65 @@ const (
...
+var runner CmdRunner = DefaultRunner{}
+
+// CmdRunner ...
+type CmdRunner interface {
+	Output(c *command.Model) (string, error)
+	Run(c *command.Model) error
+	RunWithRetry(c *command.Model) error
+}
+
+// DefaultRunner ...
```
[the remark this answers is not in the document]
lpusok: Makes sense, moved.

gitclone/git.go
```diff
@@ -28,6 +28,65 @@ const (
 	fetchFailedTag    = "fetch_failed"
 )
 
+var runner CmdRunner = DefaultRunner{}
+
+// CmdRunner ...
+type CmdRunner interface {
+	Output(c *command.Model) (string, error)
```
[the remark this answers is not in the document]
lpusok: Renamed.

gitclone/git.go
```diff
@@ -28,6 +28,65 @@ const (
 	fetchFailedTag    = "fetch_failed"
 )
 
+var runner CmdRunner = DefaultRunner{}
+
+// CmdRunner ...
+type CmdRunner interface {
```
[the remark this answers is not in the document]
lpusok: Renamed.

gitclone/git.go
lpusok: Removed extra indirection.

gitclone/gitclone_test.go
lpusok: Usually wantErr bool, but in this case makes sense, renamed.

gitclone/gitclone_test.go
lpusok: I think excessive use of constants makes the tests harder to read.

gitclone/gitclone_test.go
lpusok: Moved test cases out of the function, categorized them better.

gitclone/gitclone_test.go
lpusok: It is doing the exact same thing, except require.Equal prints out a better diff on case of failure. Do not want to introduce the testify dependency for a single test yet.

gitclone/gitclone_test.go
adborbas: I would agree with you if the actual value of repoURL and commit had an impact on the test outcome, but this is not the case, at least not in every test. By extracting them to constants you actually make the test less cluttered by only exposing values that are important for that particular test case.

gitclone/gitclone_test.go
adborbas: Do you agree to this?

gitclone/gitclone_test.go
lpusok: removed repo URL where not needed.

gitclone/gitclone_test.go
adborbas: I'll create a new PR showcasing how that would look like.

adborbas (approved):

PR #138, opened 2021-03-11
Step 618 fetch refspec

godrei:
### Checklist

### Context

Fixes:  commit push triggered git fetch does not have any refspec set.

Resolves: https://bitrise.atlassian.net/browse/STEP-618

### Changes

- remove `*string` usages, as we do not need to differentiate if the branch is specified as empty string or not specified at all
- pass branch to commit checkout and apply as refspec for fetch
- remove `_test_checkout_commit_with_other_branch` test, as it was fixing the behavior we are about to change here

### DIFF
--- a/gitclone/gitclone_test.go
+++ b/gitclone/gitclone_test.go
@@ -44,7 +44,7 @@ var testCases = [...]struct {
 			Branch: "hcnarb",
 		},
 		wantCmds: []string{
-			`git "fetch"`,
+			`git "fetch" "origin" "refs/heads/hcnarb"`,
 			`git "checkout" "76a934ae"`,
 		},
 	},
@@ -113,7 +113,7 @@ var testCases = [...]struct {
 			Branch: "hcnarb",
 		},
 		wantCmds: []string{
-			`git "fetch" "--tags"`,
+			`git "fetch" "--tags" "origin" "refs/heads/hcnarb"`,
 			`git "checkout" "76a934ae"`,
 		},
 	},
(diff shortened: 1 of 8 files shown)


gitclone/gitclone_test.go
```diff
@@ -113,7 +113,7 @@ var testCases = [...]struct {
 			Branch: "hcnarb",
 		},
 		wantCmds: []string{
-			`git "fetch" "--tags"`,
+			`git "fetch" "--tags" "origin" "refs/heads/hcnarb"`,
```
mateherber, line 116: Instead of this we may fetch only tags: https://stackoverflow.com/questions/1204190/does-git-fetch-tags-include-git-fetch

mateherber: I mean not in this scenario but when only tag is provided

mateherber: BTW I wouldn't fetch tags here only if the "all-tags" input is true

lpusok:
Can you link a succesful checkout workflow?

gitclone/gitclone_test.go
godrei: In this PR I tried to change a single behavior of the step: each fetch should have a branch as refspec set (if branch is provided) so that fetches are limited to that branch.

What about adding this information to the related ticket (Git Clone Step: introduce an input for tag fetch control), I was not aware of this 🤔

godrei:
@lpusok added a new test case (`_test_checkout_commit_on_branch`), see [build log](https://app.bitrise.io/build/f5e7c0036d5d5407#?tab=log) line 1377.

mateherber (approved):

lpusok (approved):

PR #139, opened 2021-03-11
Implement fallback to manual merge upon failing on apply

istvankovacs-bitrise:
### Checklist

### Context
Previously there was a retry mechanism in place for applying a patch file for a certain branch, however, this behavior has been removed in _v4.0.0_. Based on user reports this resulted in an increased number of Step failures, so a revamped fallback logic is implemented.

If an error occurs during applying a patch file, the Step fallbacks to the _manual merge_ strategy.

Resolves: https://bitrise.atlassian.net/browse/STEP-200

### Changes
- `PRDiffFileParams` was extended by two new parameters:
  - `PRManualMergeParam` either holds a `PRManualMergeParams` object or `nil` (in case the PR has opened to the remote repository)
  - `ForkPRManualMergeParam` either holds a `ForkPRManualMergeParams` object or `nil` (in case the PR has opened on a forked repository)
- These new parameters are populated by the time the `CheckoutPRDiffFileMethod` was selected. This is done by the `createManualMergeParams` which instantiates the corresponding object, based on the result of the `isFork` method.
- Fallback logic is implemented by the `fallbackToManualMergeOnApplyError` method, which is executed upon an error has occurred during applying the patch file.
- Some reused parts (by the fallback method) were extracted in order to avoid code duplication.

### Investigation details

See the sub-tasks of [STEP-200](https://bitrise.atlassian.net/browse/STEP-200) for a detailed resolution.

### DIFF
--- a/bitrise.yml
+++ b/bitrise.yml
@@ -32,13 +32,12 @@ workflows:
     envs:
     - CLONE_INTO_DIR: .
     - GIT_REPOSITORY_URL: https://github.com/bitrise-io/git-clone-test.git
+    before_run:
+    - audit-this-step
+    - go-tests
     steps:
     - activate-ssh-key:
         run_if: true
-    - go-list:
-    - golint:
-    - errcheck:
-    - go-test:
     after_run:
     - _test_error
     - _test_submodule
@@ -60,6 +59,13 @@ workflows:
     - _test_unrelated_histories
     - _test_hosted_git_ssh_prefix
     - test_diff_file
+
+  go-tests:
+    steps:
+    - go-list:
+    - golint:
+    - errcheck:
+    - go-test:
 
   _test_submodule:
     before_run:
@@ -686,8 +692,6 @@ workflows:
         inputs:
         - path: $STEP_TMPDIR
 
-  # ----------------------------------------------------------------
-  # --- workflow to Share this step into a Step Library
   audit-this-step:
     steps:
     - script:
--- a/bitrise.yml
+++ b/bitrise.yml
@@ -635,7 +635,7 @@ workflows:
         - pull_request_merge_branch: ""
         - branch_dest: "master"
         - branch: "master"
-        - commit: "commit"
+        - commit: "c6810e6"
         - tag: ""
         - clone_depth: "1"
         - manual_merge: "no"
--- a/bitrise.yml
+++ b/bitrise.yml
@@ -12,8 +12,8 @@ app:
       echo "GIT_CLONE_COMMIT_AUTHOR_EMAIL: ${GIT_CLONE_COMMIT_AUTHOR_EMAIL}"
       echo "GIT_CLONE_COMMIT_COMMITER_NAME: ${GIT_CLONE_COMMIT_COMMITER_NAME}"
       echo "GIT_CLONE_COMMIT_COMMITER_EMAIL: ${GIT_CLONE_COMMIT_COMMITER_EMAIL}"
-      opts:
-    is_expand: false
+    opts:
+      is_expand: false
   # define these envs in your .bitrise.secrets.yml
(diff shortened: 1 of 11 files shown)


bitrise.yml
```diff
@@ -635,7 +635,7 @@ workflows:
         - pull_request_merge_branch: ""
         - branch_dest: "master"
         - branch: "master"
-        - commit: ""
+        - commit: "commit"
```
godrei, line 638: this seems to be a strange commit hash to me, did we intentionally added this here?

gitclone/checkout.go
```diff
@@ -228,3 +233,17 @@ func selectFallbacks(checkoutStrategy CheckoutMethod, fetchOpts fetchOptions) fa
 		return nil
 	}
 }
+
+func createManualMergeParams(cfg Config) (*PRManualMergeParams, *ForkPRManualMergeParams, error) {
```
godrei, line 237: this could be simplified by using named return values, something like:

```
func createManualMergeParams(cfg Config) (prManualMergeParam *PRManualMergeParams, forkPRManualMergeParam *ForkPRManualMergeParams, err error) {
	if isFork(cfg.RepositoryURL, cfg.PRRepositoryURL) {
 		forkPRManualMergeParam, err = NewForkPRManualMergeParams(cfg.Branch, cfg.PRRepositoryURL, cfg.BranchDest)
 	} else {
 		prManualMergeParam, err = NewPRManualMergeParams(cfg.Branch, cfg.Commit, cfg.BranchDest)
 	}
       return
}
```

lpusok (approved):

gitclone/checkout_helper.go
```diff
@@ -26,6 +37,7 @@ func (t fetchOptions) IsFullDepth() bool {
 }
 
 const branchRefPrefix = "refs/heads/"
+const forkRemoteName = "fork"
```
godrei, line 40: i suggest moving this close to the `defaultRemoteName` const (gitclone.go ln 35) as they are related.

godrei: I also would rename `defaultRemoteName` to `originRemoteName`, since in the case of a fork, we do the naming based on the stored remote name (rather than which one is default or not).

gitclone/checkout_helper.go
```diff
@@ -123,6 +135,23 @@ func mergeWithCustomRetry(gitCmd git.Git, arg string, retry fallbackRetry) error
...
+func fetchAndMerge(gitCmd git.Git, fetchParam fetchParams, mergeParam mergeParams) error {
+	headBranchRef := branchRefPrefix + fetchParam.branch
+	if err := fetch(gitCmd, fetchParam.remote, &headBranchRef, fetchParam.options); err != nil {
+		return nil
+	}
+
+	return mergeWithCustomRetry(gitCmd, mergeParam.arg, mergeParam.fallback)
+}
+
+func addForkRemote(gitCmd git.Git, repoURL string) error {
```
godrei, line 147: does it worth wrapping `gitCmd.RemoteAdd` with this function?

gitclone/checkout.go
istvankovacs-bitrise: You're right!

gitclone/checkout_helper.go
istvankovacs-bitrise: I agree, not worth being in a separated function.

bitrise.yml
istvankovacs-bitrise: This is also needed based on our conversation this morning, that a commit also has to be provided.

godrei: I see, this case it would be great to specify the valid commit hash of the PR commit

godrei (approved):

PR #141, opened 2021-03-16
Merge manual merge strategies together

istvankovacs-bitrise:
https://bitrise.atlassian.net/browse/STEP-688

### Checklist

### Context
The previous intention that we should merge the two (regular and fork) manual merge strategies together was strengthened by [this](https://github.com/bitrise-steplib/steps-git-clone/pull/139) PR, since a huge amount of code duplication and simplification can be done by this change.

### Changes
- Standalone `CheckoutForkPRManualMergeMethod` checkout strategy removed.
- The `CheckoutPRManualMergeMethod` checkout strategy was extended by the parameters of the fork method, also with an attribute indicating whether the repository is a fork or not.
- The two (regular and fork) actions merged together into the same `do` method, toggled the `isFork` attribute, which is needed to execute.
- Auto-merge strategy updated to contain and use the unified manual merge as a fallback.
- Some unused helper methods removed.
- Tests updated.

### Investigation details
There was no need for further investigation.

### DIFF
--- a/gitclone/checkout.go
+++ b/gitclone/checkout.go
@@ -219,12 +219,3 @@ func selectFallbacks(checkoutStrategy CheckoutMethod, fetchOpts fetchOptions) fa
 		return nil
 	}
 }
-
-// func createManualMergeParams(cfg Config) (prManualMergeParam *PRManualMergeParams, forkPRManualMergeParam *ForkPRManualMergeParams, err error) {
-// 	if isFork(cfg.RepositoryURL, cfg.PRRepositoryURL) {
-// 		forkPRManualMergeParam, err = NewForkPRManualMergeParams(cfg.Branch, cfg.PRRepositoryURL, cfg.BranchDest)
-// 	} else {
-// 		prManualMergeParam, err = NewPRManualMergeParams(cfg.Branch, cfg.Commit, cfg.BranchDest)
-// 	}
-// 	return
-// }
--- a/gitclone/checkout.go
+++ b/gitclone/checkout.go
@@ -169,8 +169,12 @@ func createCheckoutStrategy(checkoutMethod CheckoutMethod, cfg Config, patch pat
 		}
 	case CheckoutPRManualMergeMethod:
 		{
-			isFork := isFork(cfg.RepositoryURL, cfg.PRRepositoryURL)
-			params, err := NewPRManualMergeParams(isFork, cfg.Branch, cfg.Commit, cfg.PRRepositoryURL, cfg.BranchDest)
+			prRepositoryURL := ""
+			if isFork(cfg.RepositoryURL, cfg.PRRepositoryURL) {
+				prRepositoryURL = cfg.PRRepositoryURL
+			}
+
+			params, err := NewPRManualMergeParams(cfg.Branch, cfg.Commit, prRepositoryURL, cfg.BranchDest)
 			if err != nil {
 				return nil, err
 			}
--- a/gitclone/checkout.go
+++ b/gitclone/checkout.go
@@ -92,10 +92,6 @@ func selectCheckoutMethod(cfg Config) CheckoutMethod {
 		return CheckoutPRDiffFileMethod
 	}
 
-	if isFork {
-		return CheckoutForkPRManualMergeMethod
-	}
-
 	return CheckoutPRManualMergeMethod
 }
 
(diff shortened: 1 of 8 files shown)


gitclone/checkout.go
```diff
@@ -171,20 +167,10 @@ func createCheckoutStrategy(checkoutMethod CheckoutMethod, cfg Config, patch pat
...
-
-			return checkoutForkPRManualMerge{
-				params: *params,
-			}, nil
-		}
 	case CheckoutPRManualMergeMethod:
 		{
-			params, err := NewPRManualMergeParams(cfg.Branch, cfg.Commit, cfg.BranchDest)
+			isFork := isFork(cfg.RepositoryURL, cfg.PRRepositoryURL)
+			params, err := NewPRManualMergeParams(isFork, cfg.Branch, cfg.Commit, cfg.PRRepositoryURL, cfg.BranchDest)
```
godrei, line 173: if we only set `PRRepositoryURL` in case of a fork, we could avoid the `isFork` input.

gitclone/checkout_method_pr_auto_diff.go
```diff
@@ -58,13 +55,9 @@ func (c checkoutPRDiffFile) do(gitCmd git.Git, fetchOptions fetchOptions, fallba
...
 
-		return fallbackToManualMergeOnApplyError(
-			gitCmd,
-			c.params.PRManualMergeParam,
-			c.params.ForkPRManualMergeParam,
-			fetchOptions,
-			fallback,
-		)
+		if err := c.params.PRManualMergeStrategy.do(gitCmd, fetchOptions, fallback); err != nil {
+			return fmt.Errorf("Applying patch (%s) failed: %v", c.patchFile, err)
```
godrei, line 59: `Manual merge failed: %s, err`

istvankovacs-bitrise:
You may see an example of an unable patch application for a sample repository [here](https://app.bitrise.io/build/dacd8f1a5388cf52#?tab=log).

gitclone/checkout_method_pr_manual.go
```diff
@@ -8,100 +8,72 @@ import (
...
-		return nil, NewParameterValidationError("manual PR merge checkout strategy can not be used: no base branch specified")
+
+	if isFork {
+		remoteForkBranch := fmt.Sprintf("%s/%s", forkRemoteName, headBranch)
+		return &PRManualMergeParams{
+			IsFork:      isFork,
+			HeadBranch:  headBranch,
+			MergeArg:    remoteForkBranch,
+			HeadRepoURL: forkRepoURL,
+			BaseBranch:  baseBranch,
```
godrei, line 37: might the shared properties could be set before this if, inside the if we could set the case-specific parameters, this way we could better highlight the difference

gitclone/checkout_method_pr_auto_diff.go
```diff
@@ -52,7 +52,7 @@ func (c checkoutPRDiffFile) do(gitCmd git.Git, fetchOptions fetchOptions, fallba
 	}
 
 	if err := runner.Run(gitCmd.Apply(c.patchFile)); err != nil {
-		log.Warnf("Could not apply patch (%s): %v", c.patchFile, err)
```
godrei, line 58: we did we remove the error? would not it be useful for debugging?

gitclone/checkout.go
istvankovacs-bitrise: I was not sure about only either one or the other exists. For instance, `PR - no fork - manual merge: repo is the same with different schem` test sets both the `PRMergeBranch` and the `Commit` parameters. Since the `RepositoryURL` is not passed to the param instantiation function, I thought the safest way is just to pass this info as an `isFork` parameter.
However, if we could verify that only one of them is set, and therefore we can use it to identify a fork, I'm happy to remove and simplify it.

gitclone/checkout_method_pr_auto_diff.go
istvankovacs-bitrise: I believe if the manual merge fails for some reason, it will indicate its failure (with something like this: `Manual merge failed`). My intention here was to signal that the fallback option has also failed, so there is no further way in order to apply the patch. 

I agree that the error message in question does not clearly reflect what I just mentioned here, so if you agree, I would just update this error message to express it better.

gitclone/checkout_method_pr_manual.go
istvankovacs-bitrise: Sounds reasonable!

gitclone/checkout.go
godrei: I ment somthing like this:

```
			prRepositoryURL := ""
			isFork := isFork(cfg.RepositoryURL, cfg.PRRepositoryURL)
			if isFork {
				prRepositoryURL = cfg.PRRepositoryURL
			}
			params, err := NewPRManualMergeParams(cfg.Branch, cfg.Commit, prRepositoryURL, cfg.BranchDest)
			if err != nil {
				return nil, err
			}
```

This way the presents of prRepositoryURL indicates if the PR head repo is a fork.

In the case of the `PR - no fork - manual merge: repo is the same with different scheme` test `prRepositoryURL` would remain blank.

istvankovacs-bitrise: Sounds good!

godrei (approved):

PR #143, opened 2021-03-16
Support shallow cloning submodules

adborbas:
### Checklist

- Requires MINOR [version update](https://semver.org/)

### Context

Support shallow cloning submodules to speed up cloning of repositories with submodules.

Resolves: [STEP-51](https://bitrise.atlassian.net/browse/STEP-51)

### Changes

- Updated `github.com/bitrise-io/go-utils/command/git`
- Introduced `limit_submodule_update_depth` input with default value of `yes`. If input has `yes` value that submodules will be cloned with the `--depth=1` argument.
- Create tests to cover the case.

### DIFF
--- a/step.yml
+++ b/step.yml
@@ -108,7 +108,7 @@ inputs:
   - limit_submodule_update_depth: "yes"
     opts:
       category: Clone Config
-      title: "Shalow update submodules"
+      title: "Shallow update submodules"
       summary: "Shoud submodules be shallow cloned with a history truncated to the latest revisions."
       value_options:
         - "yes"
--- a/step.yml
+++ b/step.yml
@@ -1,34 +1,34 @@
 title: "Git Clone Repository"
 summary: The Step checks out the defined repository state, optionally updates the repository submodules and exports the achieved git repository state properties.
 description: |
-   The checkout process depends on the checkout properties: the Step either checks out a repository state defined by a git commit or a git tag, or achieves a merged state of a pull / merge request.
-   The Step uses two solutions to achieve the merged state of the pull / merge request: auto merge in the case of a merge branch or diff file (provided by the Git service) and manual merge otherwise.
-   Once the desired state is checked out, the Step optionally updates the submodules. In the case of pull / merge request, the Step checks out a detach head and exports the achieved git state properties.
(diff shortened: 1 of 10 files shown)


istvankovacs-bitrise (dismissed):

step.yml
```diff
@@ -96,34 +96,42 @@ inputs:
...
     opts:
       category: Clone Config
       title: "Limit fetching to the specified number of commits"
       description: |-
         Limit fetching to the specified number of commits.
         The value should be a decimal number, for example `10`.
+  - limit_submodule_update_depth: "yes"
+    opts:
+      category: Clone Config
+      title: "Shalow update submodules"
```
lpusok, line 111: Typos, please check with techwriters.

lpusok:
Can you please link a sample workflow with a real repository, with this option applied?

adborbas:
Good catch, I've forgotten to update the e2e tests. I will added the test case for this. 👍

adborbas:
Actually, `_test_submodule` should already cover this case as both `update_submodules` and `limit_submodule_update_depth` has a default "yes" value. Do you think we are covered with that?

lpusok:
@adborbas 
>  Do you think we are covered with that?

Based on https://stackoverflow.com/questions/2144406/how-to-make-shallow-git-submodules/17692710#17692710 shallow update could fail if the referenced commit is not a branch/tag but they added fixes to check out by commit hash (which not every git server supports). Can you check that the main providers work if the referenced submodule version is not on the tip of branch/tag? In https://github.com/bitrise-io/git-clone-test unsure which commit the submodule points to.

step.yml
```diff
@@ -96,34 +96,42 @@ inputs:
...
       category: Clone Config
       title: "Limit fetching to the specified number of commits"
       description: |-
         Limit fetching to the specified number of commits.
         The value should be a decimal number, for example `10`.
+  - limit_submodule_update_depth: "yes"
+    opts:
+      category: Clone Config
+      title: "Shallow update submodules"
+      summary: "Shoud submodules be shallow cloned with a history truncated to the latest revisions."
```
lpusok, line 112: typo in Shoud

lpusok (approved):
LGTM except one typo.

lpusok (approved):

PR #146, opened 2021-03-17
Adding option to handle PRs as push events

lpusok:
### Checklist

### Context

Make PR merge optional, in case git provider allows to set the PR to be required up-to-date before merging.

Resolves: https://bitrise.atlassian.net/browse/STEP-94

### Changes

- Introduced two checkout strategies: checking out head branch and checking out a branch from a fork. The implementation is the same as a simple commit checkout.


step.yml
```diff
@@ -54,42 +54,50 @@ inputs:
...
     opts:
-      category: Clone Config
+      category: "Clone arguments"
       title: "Git Branch to clone"
       is_dont_change_value: true
   - branch_dest: "$BITRISEIO_GIT_BRANCH_DEST"
     opts:
-      category: Clone Config
+      category: "Clone arguments"
       title: "Destination git Branch, used for pull requests"
```
BanyikAnna, line 73: Is there any reason for having Branch with a capital B? I'm just curious. If not, we can revert to lower case :)

step.yml
```diff
@@ -100,19 +108,32 @@ inputs:
...
         - "yes"
         - "no"
+  - merge_pr: "yes"
+    opts:
+      category: "Checkout options"
+      title: "Merge the Pull Request source branch into target"
+      summary: "Allows to validate a temporary merged state. Optional if head branch is required to be up-to-date before merging."
+      description: |-
+        Allows disabling merging the head and base branches, 
+        in case the git provider has an option to require the Pull Request branch to be up-to-date.
```
BanyikAnna, line 131: Blocks merging the head and base branches, if the git provider has an option to require the pull request banch to be up-to-date.

I hope this is what you mean here.:)

lpusok: We do not block anything, just that this option allows to handle a PR as a push event (so no merge will occur as normally)

BanyikAnna: Disables merging the head and base branches if the the git provider has an option to require the Pull Request branch to be up-to-date.

gitclone/checkout.go
```diff
@@ -53,49 +59,100 @@ type checkoutStrategy interface {
...
+// | tag         |        |  X !|        |          |          |           |          |                   |                   |
+// | branch      |  _     |  _  |  X !   |  X       |  X       |  X        |          |                   |  X                |
+// | branchDest  |        |     |        |  X       |  X       |           |  X       |                   |                   |
+// | PRRepoURL   |        |     |        |      !   |  X !     |    !      |    !     |                   |  X !              |
+// | PRID        |        |     |        |          |          |           |    !     |                   |    !              |
+// | mergeBranch |        |     |        |          |          |  X !      |          |    !              |                   |
+// | headBranch  |        |     |        |          |          |           |          |  X !              |                   |
+// |==========================================================================================================================|
+
+func selectCheckoutMethod(cfg Config, patch patchSource) (CheckoutMethod, string) {
```
adborbas, line 74: This function is doing more than the name suggests. Instead of returning the diff file in a second parameter, we could let the caller retrieve the patch file if the function returns `CheckoutPRDiffFileMethod`. As far as I could tell in all other cases we simply return an empty string. How does that sound?

lpusok: The intention was to detect early if a diff file is unavailable and possibly fall back to an other method.

adborbas (approved):

step.yml
lpusok: no there was none.

gitclone/checkout.go
```diff
@@ -53,49 +59,100 @@ type checkoutStrategy interface {
...
 		}
 
-		return CheckoutNoneMethod
+		return CheckoutNoneMethod, ""
 	}
 
 	isFork := isFork(cfg.RepositoryURL, cfg.PRRepositoryURL)
-	isPrivateFork := isPrivate(cfg.PRRepositoryURL) && isFork
+	isPrivateSourceRepo := isPrivate(cfg.PRRepositoryURL)
+	isPrivateFork := isFork && isPrivateSourceRepo
```
godrei, line 94: It would be great to consolidate the naming.
We use `source` and `head` for PR head properties (repo, branch, etc)

gitclone/checkout.go
```diff
@@ -217,7 +291,7 @@ func selectFallbacks(checkoutStrategy CheckoutMethod, fetchOpts fetchOptions) fa
 		return nil
 	case CheckoutCommitMethod, CheckoutTagMethod:
 		return simpleUnshallow{}
-	case CheckoutPRMergeBranchMethod, CheckoutPRManualMergeMethod, CheckoutForkPRManualMergeMethod, CheckoutPRDiffFileMethod:
+	case CheckoutPRMergeBranchMethod, CheckoutPRManualMergeMethod, CheckoutPRDiffFileMethod:
```
godrei, line 294: do not we want to add fallback strategy for the 2 new checkout methods?

gitclone/checkout.go
lpusok: Updated every usage to source/destination.

gitclone/checkout.go
lpusok: Fixed.

godrei (approved):

PR #153, opened 2021-03-24
Add `sparse_directories` input to support cloning specific directories of a repo.

adborbas:
### Checklist

- Requires MINOR [version update](https://semver.org/)

### Context

Add `sparse_directories` input to support cloning specific directories of a repo.

Resolves: [STEP-626](https://bitrise.atlassian.net/browse/STEP-626)

### Changes

- `sparse_directories` can be a single directory or a multiple directories
 ```
- sparse_directories: client/android
OR
- sparse_directories: |-
                client/android
                client/common
 ```
- If `sparse_directories` is specified the `--filter=tree:0` flag will be passed to `git fetch` and `--depth` flag will be removed. More info on [--filter=tree:0](https://github.blog/2020-12-21-get-up-to-speed-with-partial-clone-and-shallow-clone/#user-content-treeless-clones)


lpusok:
Could using depth and tree filter together decrease speed? I would explicitly disable using them together if no reason to be used together.

gitclone/checkout.go
```diff
@@ -269,12 +269,31 @@ func createCheckoutStrategy(checkoutMethod CheckoutMethod, cfg Config, patchFile
...
+	return opts
+}
+
+func selectFilterTreeFetchOption(checkoutStrategy CheckoutMethod, opts fetchOptions, filterTree bool) fetchOptions {
+	if !filterTree {
+		return opts
+	}
+
+	switch checkoutStrategy {
+	case CheckoutCommitMethod, CheckoutTagMethod, CheckoutBranchMethod:
```
lpusok, line 290: Can filtertree be used in case of PRs?

adborbas: I'm not sure, will check it.

adborbas: I'm not sure if all the various PR strategies (`CheckoutPRMergeBranchMethod`, `CheckoutPRManualMergeMethod`, `CheckoutPRDiffFileMethod`) would work with this approach. My gut feeling is there could be scenarios where it could work but I'm pretty sure there would be cases where it won't. 

My suggestion would be to keep sparse-checkout for the simple cases where a commit, tag or branch should be checked out and built.

lpusok: CheckoutPRMergeBranchMethod, CheckoutPRManualMergeMethod both fetch a branch then merge an other branch, which is pretty common scenario and can be tested. I agree that CheckoutPRDiffFileMethod may not work, I am ok if we do not support it for now.

step.yml
```diff
@@ -132,6 +132,16 @@ inputs:
       value_options:
       - "yes"
       - "no"
+  - sparse_directories: ""
+    opts:
+      category: "Checkout options"
+      title: "Specify which directories to clone."
+      description: |-
+        Limit which directories should be cloned during the build. This could be useful if a repository contains multiple platforms, so called monorepo, and the build is only targeting a single platform.
```
BanyikAnna, line 140: monorepository >>monorepositories

step.yml
```diff
@@ -132,6 +132,16 @@ inputs:
...
+        For example, specifying "src/android" the step will only clone:
```
BanyikAnna, line 141: step >> Step

gitclone/checkout.go
```diff
@@ -269,12 +269,31 @@ func createCheckoutStrategy(checkoutMethod CheckoutMethod, cfg Config, patchFile
...
+	return opts
+}
+
+func selectFilterTreeFetchOption(checkoutStrategy CheckoutMethod, opts fetchOptions, filterTree bool) fetchOptions {
+	if !filterTree {
+		return opts
+	}
+
+	switch checkoutStrategy {
+	case CheckoutCommitMethod, CheckoutTagMethod, CheckoutBranchMethod, CheckoutPRMergeBranchMethod, CheckoutPRManualMergeMethod:
```
lpusok, line 290: Can you add CheckoutHeadBranchCommitMethod and CheckoutForkCommitMethod also? They check out a commit.

bitrise.yml
```diff
@@ -822,6 +823,33 @@ workflows:
           inputs:
             - dir_to_check: $CLONE_INTO_DIR
 
+  _test_sparse_checkout_commit_on_branch:
```
lpusok, line 826: Can you also add test for checking out commit cfba2b01332e31cb1568dbf3f22edce063118bae (the same as in _test_unshallow)?

lpusok (approved):

PR #163, opened 2021-06-01
Update workflows

lpusok:


bitrise.yml
```diff
@@ -1,903 +1,53 @@
...
-        echo "GIT_CLONE_COMMIT_AUTHOR_EMAIL: ${GIT_CLONE_COMMIT_AUTHOR_EMAIL}"
-        echo "GIT_CLONE_COMMIT_COMMITER_NAME: ${GIT_CLONE_COMMIT_COMMITER_NAME}"
-        echo "GIT_CLONE_COMMIT_COMMITER_EMAIL: ${GIT_CLONE_COMMIT_COMMITER_EMAIL}"
-      opts:
-        is_expand: false
-    # define these envs in your .bitrise.secrets.yml
-    - GIT_CLONE_SSH_PRIVATE_KEY: $GIT_CLONE_SSH_PRIVATE_KEY
+    - TEST_APP_URL: https://github.com/bitrise-io/git-clone-test.git
+    - TEST_APP_BRANCH: master
+    - SHOULD_ACTIVATE_SSH_KEY: true # Set to true to test with private repository
```
mateherber, line 8: I wouldn't add configurability to the sample workflow. The test app URL is public so we don't need auth support. If you want to test it you should edit the sample wf as you see fit or use the e2e steps.

bitrise.yml
```diff
@@ -1,903 +1,53 @@
...
+            - clone_into_dir: .
+            - repository_url: $TEST_APP_URL
+            - branch: $TEST_APP_BRANCH
       - script:
           inputs:
             - content: |-
                 #!/bin/env bash
-                set -ex
-                stepman audit --step-yml ./step.yml
+                echo "GIT_CLONE_COMMIT_HASH: ${GIT_CLONE_COMMIT_HASH}"
```
mateherber, line 45: I wouldn't add this "validation" script either

e2e/bitrise.yml
```diff
@@ -0,0 +1,688 @@
...
+            - repository_url: $TEST_REPO_URL
+            - clone_into_dir: $WORKDIR
+            - tag: $BITRISE_GIT_TAG
+            - branch: "test/generate-changelog"
+            - fetch_tags: "yes"
+            - update_submodules: "no"
+    after_run:
+      - _assert_changelog
+
+  _assert_changelog:
```
mateherber, line 56: I think this is a `_check_*` wf

e2e/bitrise.yml
```diff
@@ -0,0 +1,688 @@
...
+                envman unset --key GIT_CLONE_COMMIT_HASH
+                envman unset --key GIT_CLONE_COMMIT_MESSAGE_SUBJECT
+                envman unset --key GIT_CLONE_COMMIT_MESSAGE_BODY
+                envman unset --key GIT_CLONE_COMMIT_COUNT
+                envman unset --key GIT_CLONE_COMMIT_AUTHOR_NAME
+                envman unset --key GIT_CLONE_COMMIT_AUTHOR_EMAIL
+                envman unset --key GIT_CLONE_COMMIT_COMMITER_NAME
+                envman unset --key GIT_CLONE_COMMIT_COMMITER_EMAIL
+
+  _log_outputs:
```
mateherber, line 661: This is also a `_check_*` wf

e2e/bitrise.yml
```diff
@@ -0,0 +1,688 @@
...
+                fi
+                if [ ${#GIT_CLONE_COMMIT_MESSAGE_SUBJECT} -gt "20480" ]; then
+                  echo "Message subject is bigger than 20480 chars"
+                  exit 1
+                fi
+    after_run:
+      - _log_outputs
+      - _ensure_clean_git
+
+  _prepare:
```
mateherber, line 623: Probably we should have a naming convention for the `setup` kind of workflows too like `prepare` now. We might also have `teardown` kind of wfs like `_ensure_clean_git` now

I'm thinking like

```
test_*                  Test workflows
utility_*               Workflows that are used from tests
_check_*             Verify or print results
_setup_*             Prepare state for test
_teardown_*      Clean up state after test
```

mateherber (changes requested):

lpusok:
> I wouldn't add this "validation" script either 

removed.

lpusok:
> This is also a `_check_*` wf 

Updated.

e2e/bitrise.yml
mateherber: You're right. Then I think we don't need the _setup/_teardown wfs just the _run but we will have to refactor this yml to that.

```
test_*                  Test workflows (prepare env vars for _run basically)
utility_*               Workflows that are used from tests
_check_*             Verify or print results
_run                    3 phase of step execution: setup, run the step, teardown
```

e2e/bitrise.yml
```diff
@@ -0,0 +1,688 @@
...
+          inputs:
+            - repository_url: $TEST_REPO_URL
+            - clone_into_dir: $WORKDIR
+            - commit: "76a934ae80f12bb9b504bbc86f64a1d310e5db64"
+            - branch: "test/commit-messages"
+            - update_submodules: "no"
+      - script:
+          inputs:
+            - content: |-
+                #!/usr/bin/env bash
```
mateherber, line 395: this should be extracted as check wf

istvankovacs-bitrise (approved):

mateherber (approved):

PR #165, opened 2021-07-13
Export correct author information for PRs

lpusok:
### Checklist

### Context

Commit info (author, commit message ...) for PRs (if shown at all) showed the info for the merge branch, that was created during the build (temporary only).
Will use the last commit on the  souce branch for PRs.
Diff file (used by hosted Bitbucket) is not supported, as the actual commit is not provided (only the the diff and destination branch)

Resolves: https://bitrise.atlassian.net/browse/STEP-1135

### Changes

- Added **getBuildTriggerRef** function to the **checkoutStrategy** interface
- Did some refactors to extract the git ref (source branch, commit or other), where possible

### Investigation details

https://bitrise.atlassian.net/wiki/spaces/STEP/pages/1329529095/Git+Clone+Step+commit+author+investigation

https://bitrise.atlassian.net/browse/STEP-683

### DIFF
--- a/gitclone/checkout.go
+++ b/gitclone/checkout.go
@@ -51,17 +51,11 @@ func NewParameterValidationError(msg string) error {
 	return ParameterValidationError{ErrorString: msg}
 }
 
-// authorInfo ...
-type authorInfo struct {
-	isAvailable bool
-	// git revision to source auther information from (e.g. a branch or commit)
-	gitRevision string
-}
-
 // checkoutStrategy is the interface an actual checkout strategy implements
 type checkoutStrategy interface {
 	do(gitCmd git.Git, fetchOptions fetchOptions, fallback fallbackRetry) error
-	getAuthorInfo() authorInfo
+	// commitInfoRef retruns a git ref, this will be used to get commit info (for example commit author)
+	commitInfoRef() string
 }
 
 // X: required parameter
--- a/gitclone/checkout.go
+++ b/gitclone/checkout.go
@@ -54,6 +54,7 @@ func NewParameterValidationError(msg string) error {
 // checkoutStrategy is the interface an actual checkout strategy implements
 type checkoutStrategy interface {
 	do(gitCmd git.Git, fetchOptions fetchOptions, fallback fallbackRetry) error
+
 	// getBuildTriggerRef returns ref to the commit/branch/tag that triggered the build.
 	// For simple checkout strategies the returned ref will be HEAD (after running 'do').
 	// However a PR checkout strategy may create a (temporary) merge commit, so the merged state can be tested.
--- a/gitclone/checkout.go
+++ b/gitclone/checkout.go
@@ -58,8 +58,8 @@ type commitInfoRef struct {
 // checkoutStrategy is the interface an actual checkout strategy implements
 type checkoutStrategy interface {
 	do(gitCmd git.Git, fetchOptions fetchOptions, fallback fallbackRetry) error
-	// commitInfoRef retruns a git ref, this will be used to get commit info (for example commit author)
-	commitInfoRef() *commitInfoRef
+	// getCommitInfoRef returns a git ref which is used to get commit info like commit author
+	getCommitInfoRef() *commitInfoRef
 }
 
 // X: required parameter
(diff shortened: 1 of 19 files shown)


gitclone/checkout.go
```diff
@@ -54,6 +54,11 @@ func NewParameterValidationError(msg string) error {
 // checkoutStrategy is the interface an actual checkout strategy implements
 type checkoutStrategy interface {
 	do(gitCmd git.Git, fetchOptions fetchOptions, fallback fallbackRetry) error
+	// getBuildTriggerRef returns ref to the commit/branch/tag that triggered the build.
```
Bence1001, line 57: Nit: Let's leave an empty line before header comments.

gitclone/checkout.go
```diff
@@ -54,6 +54,11 @@ func NewParameterValidationError(msg string) error {
...
+	// For simple checkout strategies the returned ref will be HEAD (after running 'do').
+	// However a PR checkout strategy may create a (temporary) merge commit, so the merged state can be tested.
+	// In this case the returned ref will point to the Source branch (or a commit on the Source branch).
```
Bence1001, line 60: Would it make sense to mention manually triggered builds here?

gitclone/checkout_test.go
```diff
@@ -251,3 +256,84 @@ func Test_selectCheckoutMethod(t *testing.T) {
 		})
 	}
 }
+
+func Test_getCommitInfoRef(t *testing.T) {
```
Bence1001, line 260: Old naming here.

gitclone/gitclone.go
```diff
@@ -215,45 +276,16 @@ func Execute(cfg Config) error {
...
+	if ref := checkoutStrategy.getBuildTriggerRef(); ref != "" {
+		fmt.Println()
+		log.Infof("Exporting commit details")
+		if err := exportCommitInfo(gitCmd, ref, isPR, maxEnvLength); err != nil {
+			return newStepError("export_envs_failed", err, "Exporting envs failed")
 		}
+	} else {
+		fmt.Println()
+		log.Warnf(`Can not export commit information like commit message and author as it is not available.
+This may happen when using Bitbucket with the "Manual merge" Input set to 'yes' (using a Diff file).`)
```
Bence1001, line 288: Nit: `Input` -> `input` doesn't need capitalization if I'm not mistaken.

Bence1001 (changes requested):
Generally LGTM! I mostly have nitpicks and some questions.

gitclone/checkout.go
lpusok: From the Step perspective it does not matter if it is manually trigerred or trigerred by webhook. It can also run locally (in theory at least).
So I am unsure why that would help.

gitclone/checkout.go
lpusok: Added.

gitclone/checkout_test.go
lpusok: Fixed.

gitclone/gitclone.go
lpusok: Indded, fixed.

gitclone/checkout.go
```diff
@@ -326,3 +331,21 @@ func selectFallbacks(checkoutStrategy CheckoutMethod, fetchOpts fetchOptions) fa
...
+		CheckoutBranchMethod:
+		return false
+	case CheckoutPRMergeBranchMethod,
+		CheckoutPRDiffFileMethod,
+		CheckoutPRManualMergeMethod,
+		CheckoutHeadBranchCommitMethod,
+		CheckoutForkCommitMethod:
+		return true
+	default:
+		return true
```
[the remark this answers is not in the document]
lpusok: In can happen but unlikely. InvalidCheckoutMethos is placeholder value (so an uninitialized value is invalid by default, and so easier detected).
There are no big differences with either true or false, but as it is more likely that a PR method is added; true is a better default.

lpusok (commented):
Adressed questions.

gitclone/checkout.go
Bence1001: Makes sense. I didn't know, so that's why I asked.

lpusok: BTW Part of the complexity in the Step is caused by the fact that the webhhook handler does not send ready made checkout strategies, but a bunch of envs.

Bence1001 (approved):
LGTM! 👍

gitclone/checkout.go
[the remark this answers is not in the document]
Bence1001: We discussed this offline and decided to panic in case of an unknown checkout method.

Issue #159, opened 2021-04-30
The step should fail if there is no access to the repo


allco:
### Troubleshooting

- [x ] I've searched [discuss.bitrise.io](https://discuss.bitrise.io/) for possible solutions.
- Which version of the step is effected? 5.1.1
- Is the issue reproducible with the latest version? YES 
- Does the issue happen sporadically, or every time? EVERY TIME
- Is the issue reproducible locally by following our [local debug guide](https://devcenter.bitrise.io/troubleshooting/debugging-your-build-on-your-own-machine/)?  NOT APPLICABLE

### Issue description
This 👇 

```
$ git "remote" "add" "origin" "git@bitbucket.org:XXXXXXXXXXXXXXXXXXXXXXXXX"
Diff file unavailable: unable to download file from: https://app.bitrise.io/build/cf5e370a4a4aaf25/diff.txt?api_token=[REDACTED]. Status code: 400
May fail due to missing authentication as Pull Request opened from a private fork.
A git hosting provider head branch or a diff file is unavailable.

....

$ git "fetch" "--jobs=10" "--no-tags" "fork" "refs/heads/prxxxx_issue_test_branch"
Forbidden
fatal: Could not read from remote repository.
Please make sure you have the correct access rights
and the repository exists.
Attempt 1 failed:
Forbidden
fatal: Could not read from remote repository.
Please make sure you have the correct access rights
and the repository exists.
Retrying...
$ git "fetch" "--jobs=10" "--no-tags" "fork" "refs/heads/prxxxx_issue_test_branch"
Forbidden
fatal: Could not read from remote repository.
Please make sure you have the correct access rights
and the repository exists.
Attempt 2 failed:
Forbidden
fatal: Could not read from remote repository.
Please make sure you have the correct access rights
and the repository exists.
Retrying...
$ git "fetch" "--jobs=10" "--no-tags" "fork" "refs/heads/prxxxx_issue_test_branch"
Forbidden
fatal: Could not read from remote repository.
Please make sure you have the correct access rights
and the repository exists.
Attempt 3 failed:
Forbidden
fatal: Could not read from remote repository.
Please make sure you have the correct access rights
and the repository exists.
```

should lead the entire task to fail

Currently the task succeeds even if it is not able to actually connect to the remote repo.

Roland-Bak:
Hello there @allco 👋
[...]

PR #167, opened 2021-08-10
If error is present return it instead of success

mateherber:
### Checklist

- [✅] I've read and accepted the [Contribution Guidelines](https://github.com/bitrise-steplib/.github/blob/main/CONTRIBUTING.md)

### Context

In the manual merge case when the step fails to fetch content from the remote it didn't fail but returned with success but didn't do what it intended.

Resolves: #159 

### Changes

- Properly return an error when it occurs


Bence1001 (approved):
Good catch! 👌

lpusok (approved):

PR #170, opened 2021-09-16
Add git LFS as a dependency

ofalvai:
### Checklist

- [X] I've read and followed the [Contribution Guidelines](https://github.com/bitrise-steplib/.github/blob/main/CONTRIBUTING.md)

### Version

### Context

Git LFS is an optional dependency that some projects utilize. Normally it is preinstalled on Bitrise stacks, but a recent VM update didn't have this installed. Let's guard against this by installing the dpenedency at runtime if needed.

### Changes

Use the `deps` feature of `step.yml` to check and install Git LFS if necessary.

### Investigation details

Performance impact:

```
vagrant@prd-atl-elt-e2-xcode-13-0-210916091305-dLSbByNQ2pcr3iXsDCfw7a ~ % time brew install git-lfs
Warning: git-lfs 2.13.3 is already installed and up-to-date.
To reinstall 2.13.3, run:
  brew reinstall git-lfs
brew install git-lfs  1.11s user 1.08s system 98% cpu 2.223 total
```

```
root@prd-20-04-use1c-210916091822-mvbwsqcsrhyed4w4zvdft9:/bitrise/src# time apt install git-lfs
Reading package lists... Done
Building dependency tree
Reading state information... Done
git-lfs is already the newest version (2.13.3).
0 upgraded, 0 newly installed, 0 to remove and 1 not upgraded.

real	0m1.718s
user	0m1.389s
sys	0m0.111s
```

This was run through remote access while the build VM was executing other steps, so I think the real perf impact is even lower.

### DIFF
--- a/step.yml
+++ b/step.yml
@@ -39,6 +39,11 @@ is_requires_admin_user: false
 is_always_run: false
 is_skippable: false
 run_if: .IsCI
+deps:
+  brew:
+  - name: git-lfs
+  apt:
+  - name: git-lfs
 toolkit:
   go:
     package_name: github.com/bitrise-steplib/steps-git-clone


shams-ahmed:
@ofalvai Do we need to call `git lfs install` afterwards to make sure it's fully installed?

ofalvai:
@shams-ahmed good catch, thank you. My understanding is that this command only adds git hooks, so let's hope those are untouched in the hypothetical situation when the Git LFS binary is missing for some reason.

lpusok (approved):

PR #185, opened 2022-11-30
Log checkout metrics

ofalvai:
### Checklist

### Version

### Context

Measure and log various metrics about the step to get more visibility into speed and usage.

### Changes

- Add `go-utils/v2/analytics` and create a global tracker instance
- Measure and log fetch+checkout times: I decided to measure the two actions together because there are multiple possible "checkout strategies" based on the build triggers and the only common thing is that they do some kind of fetch+checkout. We can add more sophisticated events later if needed.
- A couple of dependency changes as a result of including `go-utils/v2`


hisaac (approved):
Looking good to me! I'm unfamiliar with how are analytics work. Is this already tracking somewhere where we can see the data? Or do we need to wait for it to be released and used for a while?

PR #186, opened 2022-12-02
Fix merge ref checkout

ofalvai:
### Checklist

### Version

### DIFF
--- a/gitclone/checkout.go
+++ b/gitclone/checkout.go
@@ -205,12 +205,12 @@ func createCheckoutStrategy(checkoutMethod CheckoutMethod, cfg Config, patchFile
 		}
 	case CheckoutPRMergeBranchMethod:
 		{
-			params, err := NewPRMergeBranchParams(cfg.PRDestBranch, cfg.PRMergeBranch)
+			params, err := NewPRMergeRefParams(cfg.PRDestBranch, cfg.PRMergeBranch)
 			if err != nil {
 				return nil, err
 			}
 
-			return checkoutPRMergeBranch{
+			return checkoutPRMergeRef{
 				params: *params,
 			}, nil
 		}
--- a/gitclone/checkout_method_pr_auto_merge_branch.go
+++ b/gitclone/checkout_method_pr_auto_merge_branch.go
@@ -7,68 +7,61 @@ import (
 	"github.com/bitrise-io/go-utils/command/git"
 )
 
-// PRMergeBranchParams are parameters to check out a Merge/Pull Request (when a merge branch is available)
-type PRMergeBranchParams struct {
+// PRMergeRefParams are parameters to check out a Merge/Pull Request's merge ref (the result of merging the 2 branches)
+// When available, the merge ref is created by the git server and passed in the webhook.
+// Using a merge ref is preferred over a manual merge because we can shallow-fetch the merge ref only.
+type PRMergeRefParams struct {
 	DestinationBranch string
-	// Merge branch contains the changes premerged by the Git provider
-	MergeBranch string
+	// MergeRef contains the changes pre-merged by the git provider (eg. pull/7/merge)
+	MergeRef string
 }
 
-// NewPRMergeBranchParams validates and returns a new PRMergeBranchParams
-func NewPRMergeBranchParams(destBranch, mergeBranch string) (*PRMergeBranchParams, error) {
+// NewPRMergeRefParams validates and returns a new PRMergeRefParams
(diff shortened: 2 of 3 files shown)


gitclone/gitclone_test.go
```diff
@@ -12,608 +12,565 @@ import (
 
 const rawCmdError = "dummy_cmd_error"
 
-var checkoutStateTestCases = [...]struct {
```
ofalvai, line 15: I just moved this list inside the function body so that GoLand recognizes each test case individually and can navigate to the failing test case. Sorry for the messy diff :(

ofalvai: There are a few changes though in a few test cases that were incorrect

gitclone/gitclone_test.go
```diff
@@ -12,608 +12,565 @@ import (
...
-			PRDestBranch:  "master",
-			ManualMerge:   true,
-			ShouldMergePR: true,
-		},
-		wantErrType: ParameterValidationError{},
-		wantCmds:    nil,
-	},
-	{
-		name: "Checkout PR - auto merge - merge branch, with depth (unshallow needed)",
-		cfg: Config{
```
ofalvai, line 534: This test case is now deleted. It doesn't make sense, with the fixed checkout strategy we don't do a merge anymore, so the shallow fetch is not a problem.

e2e/bitrise.yml
```diff
@@ -214,19 +215,6 @@ workflows:
     - _check_outputs
     - _teardown
 
-  test_checkout_pull_request_standard_branch:
```
ofalvai, line 217: I don't think this test should have ever existed. It doesn't make sense to pass a branch name as the merge ref.
I also checked our webhook processor code, it seems to be only used with GitHub events, and it uses GitHub's merge ref format: https://github.com/search?q=repo:bitrise-io/bitrise-webhooks%20PullRequestMergeBranch&type=code

gitclone/git.go
```diff
@@ -92,23 +91,6 @@ func isPrivate(repoURL string) bool {
 	return strings.HasPrefix(repoURL, "git")
 }
 
-// If incoming branch matches to pull/x/merge pattern headBranchRefs
```
ofalvai, line 95: Its only usage was the previous merge branch checkout strategy, and I don't think it's correct.

e2e/bitrise.yml
hisaac: In theory, would we need this functionality for git providers that don't provide a merge ref?

hisaac (dismissed):
Had one question, but otherwise I'd say this looks good to me 👍

e2e/bitrise.yml
ofalvai: When the merge ref is not available, the PR branch is still available (for example, `pull/7/head`) and the step is going to fetch and checkout that.

hisaac: Sounds good 👍

ofalvai: Minor correction: I learned that sometimes even the PR head ref (`pull/7/head`) is unavailable (out of the major Git providers we support, Bitbucket is the only one not exposing this), but the PR branch name is always available and the step uses that as the fallback.

In merge requests, there is no real difference between the MR head and the MR branch, but for true pull requests, the PR head points to the destination repo while the PR branch points to the fork repo.

hisaac (approved):
Looks like you'll need to merge in the recent changes on `main`, but otherwise this looks good to me 👍

godrei (approved):

nikolaykasyanov:
Hey folks, I'd like to challenge this statement:
>No breaking changes in this version yet

With 6.2.3, one of our custom tools (that boils down to `git diff <PR destination ref>:HEAD`) broke because the reference to PR destination branch was no longer available after cloning.

I'd be glad for any workaround ideas

ofalvai:
@nikolaykasyanov you are right, this change has the side effect that the step no longer fetches `<PR destination ref>` because it's not needed when checking out the `<merge ref>`.

I see two options for a workaround:

1. Set the step's `manual_merge` input to `yes`. This forces the step to ignore the merge ref and do a manual merge after fetching both branches. Note that we are planning to remove this input in the next major step version and make the merge ref fetching the default.

2. Alternatively, modify the custom tool so that it fetches `<PR destination ref>` and the PR head (such as `pull/14/head`) and diff these two refs.

nikolaykasyanov:
@ofalvai thank you for the detailed workaround suggestions.

And thanks for constantly improving Bitrise 🚀

PR #188, opened 2022-12-12
Remove `manual_merge` input, only do it as a fallback

ofalvai:
### Checklist

### Version

Requires a *MAJOR* [version update](https://semver.org/)

### Context

When the pull request merge ref is available, it's always more efficient to checkout the merge ref than doing a manual merge. In the current implementation, `manual_merge` needs to be set to `no` to do that (default value is `yes`).

There is no other usage of this step input, it's probably a leftover from a previous state of the step.

### Changes

- Remove `manual_merge` input and use the merge ref or the patch file checkout options if they are available. Only do a manual merge if none of them are available.
- Adjust test cases, clean up no longer relevant cases

With this change, the step will do a merge ref checkout (if it's provided in the webhook) even with the default input values. 

### Decisions

Adjusting the clone depth is out of scope for this PR, that will come in a follow-up PR.


hisaac (approved):
👌

PR #190, opened 2022-12-13
Refactor output exports

ofalvai:
### Checklist

### Version

Requires no [version update](https://semver.org/)

### Context

Clean up the code that exports step outputs and gradually replace v1 go utils with the v2 equivalents.

### Changes

- Move output export logic to a separate `output.go` file. It's mostly copy-paste, but I made some small style improvements with new structs and renames.
- Make the error message more accurate that is printed when output export is not possible


gitclone/gitclone.go
```diff
@@ -220,14 +132,7 @@ func setupSparseCheckout(gitCmd git.Git, sparseDirectories []string) error {
 func Execute(cfg Config) error {
 	defer tracker.wait()
 
-	maxEnvLength, err := getMaxEnvLength()
```
ofalvai, line 223: Moved this inside the functions in `output.go`. The error is still returned, but will be nested in a `export_envs_failed` error (see L213 in the new version). I don't think it's going to cause any issues, this error happens very rarely anyway.

gitclone/gitclone.go
```diff
@@ -293,16 +198,19 @@ func Execute(cfg Config) error {
...
-	if ref := checkoutStrategy.getBuildTriggerRef(); ref != "" {
-		fmt.Println()
-		logger.Infof("Exporting commit details")
-		if err := exportCommitInfo(gitCmd, ref, isPR, maxEnvLength); err != nil {
-			return newStepError("export_envs_failed", err, "Exporting envs failed")
-		}
-	} else {
-		fmt.Println()
-		logger.Warnf(`Can not export commit information like commit message and author as it is not available.
-This may happen when using Bitbucket with the "Manual merge" input set to 'yes' (using a Diff file).`)
```
ofalvai, line 305: This looks like a suggestion, but it's not, because even if manual_merge was set to "yes", the step decided to override it and fetch the PR diff from Bitbucket directly. So there is no real solution to this problem with step inputs, the only workaround AFAIK is to rely on the env vars that are based on the webhook data.

hisaac (commented):
Code looks good to me. How would you feel about adding some unit tests to `output.go` as long as you're in here?

ofalvai:
@hisaac good idea, I added some tests for the main business logic. Originally I wanted to test the whole function, but I had soo much trouble mocking all the dependencies that I gave up. I think ultimately it's cleaner to only test the business logic, so I don't mind it anymore, but let me know what you think.

hisaac (approved):
Tests look good, thanks!

PR #191, opened 2022-12-13
Add method for determining the ideal default clone depth based if the…

hisaac:
### Checklist

### Version

Version update will be done when merging `next` into `master`.

### Context

To help speed up git clones, we want to provide more reasonable defaults for the clone depth. Most clones will now be shallow clones.

### Changes

- Clone depth is now determined based on the `CheckoutMethod` (shallow clone by default, but a depth of 50 for `CheckoutPRManualMergeMethod`). If the user specifies a value for the `clone_depth` input, we will use that value instead.
- Updated tests to match updates.
- Fixed a couple typos.

### DIFF
--- a/e2e/bitrise.yml
+++ b/e2e/bitrise.yml
@@ -521,7 +521,7 @@ workflows:
             check "GIT_CLONE_COMMIT_HASH" "76a934ae80f12bb9b504bbc86f64a1d310e5db64"
             check "GIT_CLONE_COMMIT_MESSAGE_SUBJECT" "Sample commit message subject"
             check "GIT_CLONE_COMMIT_MESSAGE_BODY" "Sample commit message body"
-            check "GIT_CLONE_COMMIT_COUNT" "25"
+            check "GIT_CLONE_COMMIT_COUNT" "1"
             check "GIT_CLONE_COMMIT_AUTHOR_NAME" "Krisztian Dobmayer"
             check "GIT_CLONE_COMMIT_AUTHOR_EMAIL" "krisztian.dobmayer@bitrise.io"
             check "GIT_CLONE_COMMIT_COMMITTER_NAME" "Krisztian Dobmayer"
--- a/e2e/bitrise.yml
+++ b/e2e/bitrise.yml
@@ -22,6 +22,7 @@ workflows:
     - BRANCH: test/generate-changelog
     - FETCH_TAGS: "yes"
     - UPDATE_SUBMODULES: "no"
+    - CLONE_DEPTH: 50
     before_run:
     - _setup
     after_run:
@@ -36,6 +37,7 @@ workflows:
     - BRANCH: test/generate-changelog
     - FETCH_TAGS: "yes"
     - UPDATE_SUBMODULES: "no"
+    - CLONE_DEPTH: 50
     before_run:
     - _setup
     after_run:
@@ -480,7 +482,7 @@ workflows:
         inputs:
         - content: |-
             #!/bin/env bash
-            set -ex
+            set -e
 
             EXPECTED_CHANGELOG="* [a409478] Add newline to the description.
             * [b002ab7] Add repository description.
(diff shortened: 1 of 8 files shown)


gitclone/gitclone.go
```diff
@@ -39,11 +39,11 @@ type Config struct {
...
 const (
-	trimEnding              = "..."
-	originRemoteName        = "origin"
-	forkRemoteName          = "fork"
-	updateSubmodelFailedTag = "update_submodule_failed"
-	sparseCheckoutFailedTag = "sparse_checkout_failed"
+	trimEnding               = "..."
+	originRemoteName         = "origin"
+	forkRemoteName           = "fork"
+	updateSubmoduleFailedTag = "update_submodule_failed"
```
hisaac, line 45: Just fixing a typo I noticed

step.yml
```diff
@@ -131,7 +131,7 @@ inputs:
     value_options:
     - "yes"
     - "no"
-- clone_depth:
+- clone_depth: -1
```
hisaac, line 134: @ofalvai I'm not sure how best to handle the situation when a user doesn't provide a value for `clone_depth`. From what I've gathered, if no value is specified here in the yaml, it will default to a value of `0` when we parse the config. The problem is, `0` is a valid value that someone could specify, so there'd be no way to determine if it was set to `0` by the user, or if it was set to that because the user provided no value. Unless you know of a way to do that?

Setting this to `-1` works sort of, but it's not really ideal.

gitclone/checkout.go
```diff
@@ -333,6 +331,26 @@ func selectFilterTreeFetchOption(method CheckoutMethod, opts fetchOptions, filte
 	}
 }
 
+func idealDefaultCloneDepth(method CheckoutMethod) int {
+	const defaultCloneDepth = 50
+
+	switch method {
+	case CheckoutNoneMethod,
+		CheckoutPRMergeBranchMethod,
+		CheckoutPRManualMergeMethod,
```
ofalvai, line 340: I don't think it's correct. Manual merging requires the commit history, this is the case that gives the "refusing to merge unrelated histories" on shallow checkouts.

step.yml
hisaac: _DOH_. I was assuming `--depth=0` was the shallow clone value for some reason, but I didn't actually try it 🤦. Never mind then, using `0` as the default value makes perfect sense. 👍 I'll update things.

gitclone/checkout.go
```diff
@@ -333,6 +331,26 @@ func selectFilterTreeFetchOption(method CheckoutMethod, opts fetchOptions, filte
...
+		CheckoutPRDiffFileMethod:
+		return 1
+	case CheckoutCommitMethod,
```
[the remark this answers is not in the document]
hisaac: Yeah, you're thinking makes sense to me. And the default value is easily overridden if a user specifically wants to clone more for some reason.

Re: `generate-changelog` - we'd probably need to update that to specify a clone depth value then, yeah?

gitclone/gitclone_test.go
```diff
@@ -87,45 +87,45 @@ func Test_checkoutState(t *testing.T) {
 			},
 		},
 		{
-			name: "Checkout tag, branch specifed",
+			name: "Checkout tag, branch specified",
```
hisaac, line 90: Fixed another typo

e2e/bitrise.yml
```diff
@@ -22,6 +22,7 @@ workflows:
     - BRANCH: test/generate-changelog
     - FETCH_TAGS: "yes"
     - UPDATE_SUBMODULES: "no"
+    - CLONE_DEPTH: 50
```
ofalvai, line 25: nit: How about using a value other than `50`? Just to test that it indeed applies the custom value and not the constant baked into the step 😄

ofalvai (approved):
LGTM, just don't forget to update the PR description!

image

e2e/bitrise.yml
hisaac: Good call, will do

- [ ] Update `CLONE_DEPTH` in e2e test to verify it's actually working

hisaac: Fixed in https://github.com/bitrise-steplib/steps-git-clone/pull/191/commits/5b99c3e1c3b8311939cd6c06a09c7ad6efd56339 👍

ofalvai (approved):

hisaac:
@ofalvai Thanks for the heads up on the PR description. Forgot about that.

Issue #89, opened 2019-03-01
Incorrect commit author name exported


chibatching:
On our build for pull request, git-clone step exports incorrect commit author's email and name.

```
+------------------------------------------------------------------------------+

| (1) git-clone                                                                |
+------------------------------------------------------------------------------+
| id: git-clone                                                                |
| version: 4.0.14                                                              |
| collection: https://github.com/bitrise-io/bitrise-steplib.git                |
| toolkit: go                                                                  |
| time: 2019-02-26T23:50:29-08:00                                              |
+------------------------------------------------------------------------------+
|                                                                              |
config:
- RepositoryURL: git@github.com:tokubai/ios-bargain.git
- CloneIntoDir: /Users/vagrant/git
- Commit: 083b961193f657666a94466c4d2b1cce90b3369f
- Tag: 
- Branch: fix_launch_from_notification
- BranchDest: master
- PRID: 3689
- PRRepositoryURL: git@github.com:tokubai/ios-bargain.git
- PRMergeBranch: pull/3689/merge
- ResetRepository: false
- CloneDepth: 0
- BuildURL: https://app.bitrise.io/build/6d0a7671c2b1cdc0
- BuildAPIToken: [REDACTED]
- UpdateSubmodules: true
- ManualMerge: true
git "init"
Initialized empty Git repository in /Users/vagrant/git/.git/
git "remote" "add" "origin" "git@github.com:tokubai/ios-bargain.git"
git "fetch" "origin" "master"
Warning: Permanently added 'github.com,192.30.253.112' (RSA) to the list of known hosts.
From github.com:tokubai/ios-bargain
 * branch                master     -> FETCH_HEAD
 * [new branch]          master     -> origin/master
git "checkout" "master"
Already on 'master'
Branch 'master' set up to track remote branch 'master' from 'origin'.
git "merge" "origin/master"
Already up to date.
commit hash: 190060f35e606915ab34da72e8362a0874915577
git "fetch" "origin" "fix_launch_from_notification"
Warning: Permanently added 'github.com,192.30.253.113' (RSA) to the list of known hosts.
From github.com:tokubai/ios-bargain
 * branch                fix_launch_from_notification -> FETCH_HEAD
 * [new branch]          fix_launch_from_notification -> origin/fix_launch_from_notification
git "merge" "083b961193f657666a94466c4d2b1cce90b3369f"
Merge made by the 'recursive' strategy.
[...]

PR #193, opened 2022-12-15
Use PR head ref in output export

ofalvai:
### Checklist

### Version

### Context

With #186, the step checks out the PR merge ref and uses that state to export outputs, such as the commit hash and commit message. This might be confusing for users (previous discussions: #89, #165), it makes more sense to use the PR head ref for output exports.

### Changes

- Fetch the PR head ref in addition to the PR merge ref. We do it with the same fetch params, so this should be a shallow fetch in most cases.
- Use the local head ref for exporting outputs
- Remove an unused parameter from the checkout method

### DIFF
--- a/gitclone/gitclone_test.go
+++ b/gitclone/gitclone_test.go
@@ -142,6 +142,7 @@ func Test_checkoutState(t *testing.T) {
 			},
 			wantCmds: []string{
 				`git "fetch" "--jobs=10" "--depth=1" "--no-tags" "--no-recurse-submodules" "origin" "refs/pull/5/merge:refs/remotes/pull/5/merge"`,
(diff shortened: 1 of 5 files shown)


gitclone/gitclone_test.go
```diff
@@ -152,10 +153,7 @@ func Test_checkoutState(t *testing.T) {
 				PRMergeBranch: "pr_test",
 				ShouldMergePR: true,
 			},
-			wantCmds: []string{
-				`git "fetch" "--jobs=10" "--no-tags" "--no-recurse-submodules" "origin" "refs/pr_test:refs/remotes/pr_test"`,
-				`git "checkout" "refs/remotes/pr_test"`,
-			},
+			wantErrType: ParameterValidationError{},
```
ofalvai, line 156: This test case is a little weird, I don't think it's valid that a build is triggered with a merge ref but without a head ref, so I changed the test to expect an error.

hisaac (approved):
LGTM. Good thinking.

PR #200, opened 2023-01-09
Clean up unused stuff in E2E tests

ofalvai:
### Checklist

- [ ] I've read and followed the [Contribution Guidelines](https://github.com/bitrise-steplib/.github/blob/main/CONTRIBUTING.md)
- [ ] `step.yml` and `README.md` is updated with the changes (if needed)

### Version

Requires a *MAJOR/MINOR/PATCH* [version update](https://semver.org/)

### Context

After removing some inputs in #196 and #188, it's time to clean up those in E2E tests too.

### Changes

- Remove unused step inputs
- Remove one useless E2E test that was testing the incorrect logic we fixed in #186

PR #201, opened 2023-01-09
Add option for unlimited clone depth

### LINKED ISSUE #199: Disabling shallow clone is no longer supported in version 7.0.0
### Troubleshooting

- [X] I've searched [discuss.bitrise.io](https://discuss.bitrise.io/) for possible solutions.
- Which version of the step is effected? **7.0.0**
- Is the issue reproducible with the latest version? **YES**
- Does the issue happen sporadically, or every time? **EVERY TIME**

### Useful information

- [FAQ](https://devcenter.bitrise.io/faq/faq-index/)
- [devcenter.bitrise.io](https://devcenter.bitrise.io/)
- [Shallow clone by default release notes](https://github.com/bitrise-steplib/steps-git-clone/releases/tag/7.0.0)

### Issue description

Disabling shallow clone is no longer supported. Previously, a `clone_depth` of `0` was disabling shallow clone.

### Steps to reproduce

1. Under `Clone options`, set a `Clone depth` of `0`
2. Run the step

**Expected outcome**: `git fetch` uses no depth
**Actual outcome**: `git fetch` uses a depth of `1`

ofalvai:
### Checklist

### Version

### Context

Fixes #199

### Changes

Treat the `-1` special value of `clone_depth` as the unlimited option and don't add any `--depth=N` flag to the fetch command in this case.

### Decisions

The value `0` is already taken by the default logic that tries to be smart (based on the checkout method) and either use `--depth=1` or `--depth=50` for manual merge.


ofalvai:
Current dependencies on/for this PR:
* master
  * **PR #200** <a href="https://app.graphite.dev/github/pr/bitrise-steplib/steps-git-clone/200" target="_blank">Graphite</a> 
    * **PR #201** <a href="https://app.graphite.dev/github/pr/bitrise-steplib/steps-git-clone/201" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/bitrise-steplib/steps-git-clone/201?utm_source=stack-comment).

gitclone/checkout.go
```diff
@@ -294,7 +294,14 @@ func createCheckoutStrategy(checkoutMethod CheckoutMethod, cfg Config, patchFile
 }
 
 func selectFetchOptions(method CheckoutMethod, cloneDepth int, fetchTags, fetchSubmodules bool, filterTree bool) fetchOptions {
+	// If cloneDepth is 0, that means the user did not set a value for it,
```
ofalvai, line 297: I moved this code block over from `gitclone.go` because this way it can be part of tests of `selectFetchOptions()`

step.yml
```diff
@@ -83,6 +83,7 @@ inputs:
       - A Step in the workflow reads the commit history in order to generate a changelog
       - A Step in the workflow runs a git diff against a previous commit
 
+      Use the value `-1` to disable the depth limit completely and fetch the entire repo history.
```
tothszabi, line 86: As I was reading your code I saw that in the `fetchConfig` struct we have two related properties:
```
struct fetchConfig {
    ...
    limitDepth bool
    depth         int
    ...
}
```
which made me think if instead having `-1` carry a special meaning we could have a separate field similar to the `limitDepth` variable in the example above. So it is just a bool field but maybe has a more precise meaning that adding `-1` value handling into the existing `clone_depth` field? 

I am not fully sure in it yet. I think I like the separate field better but just want to ask your opinion.

ofalvai: I was contemplating the same, but I ended up with the -1 value because:

- There are already too many step inputs 😄 
- Disabling the depth limit is not something we recommend (for performance and network utilization reasons), so I'd rather not "advertise" it with the same weight as other step inputs. But if someone really needs the full repo history in CI, they can set this input to `-1`.

I can change my mind, but these are the reasons why I decided to represent the unlimited option in the same step input.

Grayson (approved):

PR #202, opened 2023-01-20
Refactor to ideal step architecture

vshah23:
### Checklist

### Version

### Context

Ongoing effort to refactor steps to ideal architecture.

### Changes

A Step package was introduced and it utilizes the existing package while implementing the standard Step interface.

### Investigation details

This Step already had most of its functionality encapsulated into a different package instead of everything in `main.go`.

The project scanner step also relies on this step as a dependency. Here is the PR for this change:
https://github.com/bitrise-steplib/steps-project-scanner/pull/128


step/step.go
```diff
@@ -0,0 +1,113 @@
...
+	PRHeadBranch          string `env:"pull_request_head_branch"`
+
+	ResetRepository bool   `env:"reset_repository,opt[Yes,No]"`
+	BuildURL        string `env:"build_url"`
+	BuildAPIToken   string `env:"build_api_token"`
+}
+
+// Config is the git clone step configuration
+type Config struct {
+	Input
```
ofalvai, line 38: Why do we have a `Config` struct here if there is another struct in the `gitclone` package?

step/step.go
```diff
@@ -0,0 +1,113 @@
...
+	return cfg, nil
+}
+
+func (g GitCloneStep) Run(cfg Config) (gitclone.CheckoutStateResult, error) {
+	gitcloneCfg := convertConfig(cfg)
+	cloner := gitclone.NewGitCloner(g.logger, g.tracker, g.cmdFactory)
+	return cloner.CheckoutState(gitcloneCfg)
+}
+
+func (g GitCloneStep) ExportOutputs(runResult gitclone.CheckoutStateResult) error {
```
ofalvai, line 73: This `CheckoutStateResult` seems a bit smelly to me. If my understanding is correct, you introduced it in order to expose a public type from the `gitclone` package so that we can pass the info between `Run()` and `ExportOutputs()` here. I see a couple of problems with it:

1. A few internal types had to be converted to public because they are now part of the "result" and `step.go` needs to access them.
2. The `GitCmd` within the result is especially smelly, since it's not even a result, but I see that in line 84 it is used in a constructor. This feels like unnecessary ping-pong between the `gitclone` and this `step` package.

Can we maybe avoid "uplifting" these types to the `step` package and let the `gitclone` package handle them somehow? I feel like that would be cleaner and involve less boilerplate code.

ofalvai:
@vshah23 you might not like this, but did you know that the project scanner also depends on the `gitclone` package in this repo? 😅 I just learned it last week.

https://github.com/bitrise-steplib/steps-project-scanner/blob/master/go.mod#LL10C37-L10C37

I think your changes will be fine in the other repo, but please double-check it before merging this PR

step/step.go
vshah23: Hmm, yeah, it's redundant, but we wanted to conform to the ideal architecture while treating most of the existing code as a library for the "step" to call. This is to minimize the changes we make in this PR due to this step being so critical for all of our users.

Therefore we left the other struct in the `gitclone` package and just mapped this config to that one since this config is the one that gets parsed.

step/step.go
vshah23: Please check these commits:
8d881cb
94bb2a7

ofalvai: LGTM now!

ofalvai (dismissed):

godrei (dismissed):

godrei (approved):

PR #203, opened 2023-02-06
Revert to 6.2.2

vshah23:
### Checklist

### Version

### Context

Users were experiencing an issue where upon checking out the merge ref for a PR the merge ref was referencing an old commit which is incorrect.

### Changes

Copied the code from the version 6.2.2 release and kept the changes for `GIT_CLONE_COMMIT_COMMITTER_NAME` and `GIT_CLONE_COMMIT_COMMITTER_EMAIL`.


godrei (approved):

vincentisambart:
This is not a minor version change.
When migrating from 6.x to 7.x the behavior was changed so in some cases you could the explicit depth, and `manual_merge` was removed. With bringing back the behavior from 6.x, people who have changed their settings for 7.x now have to change them once again. And a lot of people do not check the changelog of minor version changes. Especially one with as little explanation as a cryptic "Revert to 6.2.2 by @vshah23 in #203"

ilockett:
> This is not a minor version change. When migrating from 6.x to 7.x the behavior was changed so in some cases you could remove the explicit clone depth, and `manual_merge` was removed. With bringing back the behavior from 6.x, people who have changed their settings for 7.x now have to change them once again. And a lot of people do not check the changelog of minor version changes. Especially one with as little explanation as a cryptic "Revert to 6.2.2 by @vshah23 in #203"

Agreed, we had clone depth set to -1 as per the previous instructions, and this has now just started failing with `fatal: depth -1 is not a positive number`

PR #208, opened 2023-02-24
Add support for pending merge refs

ofalvai:
### Checklist

### Version

### DIFF
--- a/gitclone/bitriseapi/pr_merge_ref.go
+++ b/gitclone/bitriseapi/pr_merge_ref.go
@@ -1,0 +1,98 @@
+package bitriseapi
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"net/http"
+	"time"
+
+	"github.com/bitrise-io/go-utils/retry"
+	"github.com/bitrise-io/go-utils/v2/log"
+	"github.com/hashicorp/go-retryablehttp"
+)
+
+var logger = log.NewLogger()
+var ErrServiceAccountAuth = errors.New(`authentication error: Bitrise can't connect to the git server to check the freshness of the merge ref.
+Check the Service Credential User in App Settings > Integrations`)
+
+type MergeRefChecker interface {
+	IsMergeRefUpToDate(ref string) (bool, error)
+}
+
+func NewMergeRefChecker(buildURL string, apiToken string, client *retryablehttp.Client) MergeRefChecker {
+	return apiMergeRefChecker{
+		buildURL: buildURL,
+		apiToken: apiToken,
+		client:   client,
+	}
+}
+
+type apiMergeRefChecker struct {
+	buildURL string
+	apiToken string
+	client   *retryablehttp.Client
+}
+
+type mergeRefResponse struct {
+	Status string `json:"status"`
+}
+
+type mergeRefFetcher func(attempt uint) (mergeRefResponse, error)
+
+func (c apiMergeRefChecker) IsMergeRefUpToDate(ref string) (bool, error) {
+	return doPoll(c.fetchMergeRef, time.Second*2)
+}
+
+func doPoll(fetcher mergeRefFetcher, retryWaitTime time.Duration) (bool, error) {
+	isUpToDate := false
+	pollAction := func(attempt uint) (err error, shouldAbort bool) {
+		resp, err := fetcher(attempt)
+		if err != nil {
+			logger.Warnf("Error while checking merge ref: %s", err)
+			logger.Warnf("Retrying request...")
+			return err, false
+		}
+		switch resp.Status {
+		case "up-to-date":
+			isUpToDate = true
+			return nil, true
+		case "auth_error": // TODO
+			return ErrServiceAccountAuth, true
+		case "pending":
+			logger.Warnf("Attempt %d: not up-to-date yet", attempt)
+			return fmt.Errorf("pending"), false
+		default:
+			return fmt.Errorf("unknown status: %s", resp.Status), false
+		}
+	}
+
+	err := retry.Times(5).Wait(retryWaitTime).TryWithAbort(pollAction)
+	if err != nil {
+		return false, err
+	}
+	return isUpToDate, nil
+}
+
+func (c apiMergeRefChecker) fetchMergeRef(attempt uint) (mergeRefResponse, error) {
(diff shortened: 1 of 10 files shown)


gitclone/checkout_method_pr_auto_diff.go
```diff
@@ -67,24 +62,3 @@ func (c checkoutPRDiffFile) do(gitCmd git.Git, fetchOptions fetchOptions, fallba
 func (c checkoutPRDiffFile) getBuildTriggerRef() string {
 	return ""
 }
-
```
ofalvai, line 70: Moved to `bitriseapi` package

e2e/bitrise.yml
```diff
@@ -374,6 +374,7 @@ workflows:
             envman add --key TEST_DIFF_REPO_URL --value ${GIT_DIR}
             envman add --key TEST_DIFF_PR_REPOSITORY_URL --value ${PR_GIT_DIR}
             envman add --key BITRISE_BUILD_URL --value "file://${DIFF_DIR}"
+            envman add --key BITRISE_BUILD_API_TOKEN --value "dummy token"
```
ofalvai, line 377: `BITRISE_BUILD_URL` and `BITRISE_BUILD_API_TOKEN` should both be defined or none of them

gitclone/tracker/tracker.go
```diff
@@ -1,4 +1,4 @@
-package gitclone
+package tracker
```
ofalvai, line 1: I had to move this to its own subpackage because I introduced an import cycle (by using the tracker not only in `gitclone`, but in `gitclone/bitriseapi` too)

gitclone/gitclone.go
```diff
@@ -35,25 +37,28 @@ type Config struct {
...
 	PRDestBranch          string
 	PRSourceRepositoryURL string
-	PRMergeBranch         string
+	PRMergeRef            string
+	PRUnverifiedMergeRef  string
 	PRHeadBranch          string
 
 	ResetRepository bool
-	BuildURL        string
-	BuildAPIToken   string
```
ofalvai, line 43: These are no longer needed here in the `Config` struct because we have instances of `bitriseapi.PatchSource` and `bitriseapi.MergeRefChecker` that contain these params.

gitclone/bitriseapi/pr_merge_ref.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+func doPoll(fetcher mergeRefFetcher, retryWaitTime time.Duration, logger log.Logger) (bool, uint, error) {
+	isUpToDate := false
+	attempts := uint(0)
+	pollAction := func(attempt uint) (err error, shouldAbort bool) {
+		resp, err := fetcher(attempt)
+		attempts = attempt + 1 // attempt is 0-indexed
+		if err != nil {
+			logger.Warnf("Error while checking merge ref: %s", err)
```
godrei, line 69: Do we need this verbose logging by default around this functionality?

gitclone/bitriseapi/pr_merge_ref.go
```diff
@@ -0,0 +1,118 @@
...
+		case "pending":
+			logger.Warnf("Attempt %d: not up-to-date yet", attempts)
+			return fmt.Errorf("pending"), false
+		default:
+			logger.Warnf("Attempt %d: unknown status: %s", attempts, resp.Status)
+			return fmt.Errorf("unknown status: %s", resp.Status), false
+		}
+	}
+
+	err := retry.Times(5).Wait(retryWaitTime).TryWithAbort(pollAction)
```
godrei, line 89: The number of retries (5 in this case), could come as a function parameter, just like `retryWaitTime`.
I would also create a const for the default number of attempts and waittime, to make them more visible.

godrei (commented):
I like it so far

gitclone/bitriseapi/pr_merge_ref.go
ofalvai: Good idea, fixed in latest commit

gitclone/bitriseapi/pr_patch.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+}
+
+type apiPatchSource struct {
+	buildURL string
+	apiToken string
+}
+
+func (s apiPatchSource) GetPRPatch() (string, error) {
+	if s.buildURL == "" {
```
godrei, line 31: Same as above shouldn't we do the validation in the constructor?

gitclone/bitriseapi/pr_merge_ref.go
```diff
@@ -0,0 +1,136 @@
...
+	return isUpToDate, err
+}
+
+func doPoll(fetcher mergeRefFetcher, maxAttemptCount uint, retryWaitTime time.Duration, logger log.Logger) (bool, uint, error) {
+	isUpToDate := false
+	attempts := uint(0)
+	pollAction := func(attempt uint) (err error, shouldAbort bool) {
+		resp, err := fetcher(attempt)
+		attempts = attempt + 1 // attempt is 0-indexed
+		if err != nil {
```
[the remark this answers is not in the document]
ofalvai: I think it's more accurate this way. We should count a failed request as an attempt too.

gitclone/bitriseapi/pr_merge_ref.go
```diff
@@ -0,0 +1,136 @@
...
+
+type mergeRefResponse struct {
+	Status string `json:"status"`
+	Error  string `json:"error_msg"`
+}
+
+type mergeRefFetcher func(attempt uint) (mergeRefResponse, error)
+
+func (c apiMergeRefChecker) IsMergeRefUpToDate(ref string) (bool, error) {
+	if c.buildURL == "" {
```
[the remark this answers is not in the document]
ofalvai: I made this change because it seemed like a good idea, then I realized why we don't do this 😅 

This `MergeRefChecker` is instantiated at the beginning of the main execution, so handling the constructor errors would happen there and would abort the entire step run. This is exactly what happened in the E2E tests after my changes, a perfectly fine test failed because it didn't mock the API URL and token, but there is no API communication in that test case.

Since this merge ref checker is just one of the many code paths and we handle its errors with a fallback anyway, I think it's better to leave it as-is, I don't want to make the local step execution (or E2E execution) require an API token.

PR #228, opened 2024-12-11
Force-checkout remote branch to fix unrelated histories

ofalvai:
### Checklist

### Version

### Context

Git clones can fail in persistent environments (bare-metal runners) when the build is triggered by a push event. Imagine this sequence:
1. Build 1 runs with the default step settings. It fetches `origin/main` (because it's not a PR build) with `--depth=1` (default step setting).
2. New commits are pushed to `origin/main`
3. Build 2 runs: it fetches `origin/main` again with `--depth=1`, but when it tries to checkout and merge the local branch, it fails with the `refusing to merge unrelated histories` error.

An actual step log:

```
$ git "init"
Reinitialized existing Git repository in /home/bitrise-admin/ca71a338453a42a3/git/.git/
$ git "config" "gc.auto" "0"
$ git "fetch" "--jobs=10" "--depth=1" "--no-tags" "origin" "refs/heads/master"
From github.com:bitrise-io/ubuntu-android
 * branch            master     -> FETCH_HEAD
 + d0413b1...db22ecf master     -> origin/master  (forced update)
$ git "checkout" "master"
Switched to branch 'master'
Your branch and 'origin/master' have diverged,
and have 1 and 1 different commits each, respectively.
  (use "git pull" to merge the remote branch into yours)
$ git "merge" "origin/master"
fatal: refusing to merge unrelated histories
Checkout strategy used: gitclone.checkoutBranch
Failed to execute Step:
  updating branch (master) failed:
    fatal: refusing to merge unrelated histories
```

### Changes

Branch checkout was implemented using a `fetch` + `checkout` + `merge remote local` sequence. This doesn't work in persistent envs because there is a leftover local branch from the previous run with a single commit (because of shallow fetching). The new remote-tracking branch also has a single commit, and there is no common ancestor of these two commits, so the merge command fails.

This PR changes the implementation of our branch checkout so that any previous local branch state is overwritten, sort of like a `git pull --force`.

The end result remains the same:
- The branch is checked out
- It is not behind origin
- If the local branch already exists, it's overwritten
- [new] If the local and remote branches diverged, the remote state wins

### DIFF
--- a/gitclone/checkout_helper.go
+++ b/gitclone/checkout_helper.go
@@ -114,9 +114,9 @@ func forceCheckoutRemoteBranch(gitCmd git.Git, remote string, branchRef string, 
 	if err != nil {
 		return handleCheckoutError(
 			listBranches(gitCmd),
-			fetchFailedTag,
+			checkoutFailedTag,
 			err,
-			"Fetching repository has failed",
+			"Checkout has failed",
 			branch,
 		)
 	}
--- a/gitclone/checkout_helper.go
+++ b/gitclone/checkout_helper.go
@@ -112,7 +112,7 @@ func forceCheckoutRemoteBranch(gitCmd git.Git, remote string, branchRef string, 
 	// so the next run would see unrelated histories after shallow-fetching another single commit.
 	out, err := runner.RunForOutput(command.New("git", "checkout", "-B", branch, remoteBranch))
 	if err != nil {
-		fmt.Printf(out)
+		fmt.Println(out)
 		return fmt.Errorf("checkout remote-tracking branch %s: %w", remoteBranch, err)
 	}
 
--- a/gitclone/checkout_helper.go
+++ b/gitclone/checkout_helper.go
@@ -110,10 +110,15 @@ func forceCheckoutRemoteBranch(gitCmd git.Git, remote string, branchRef string, 
 	// -B: create the branch if it doesn't exist, reset if it does
 	// The latter is important in persistent environments because shallow-fetching only fetches 1 commit,
 	// so the next run would see unrelated histories after shallow-fetching another single commit.
-	out, err := runner.RunForOutput(command.New("git", "checkout", "-B", branch, remoteBranch))
+	err := runner.Run(gitCmd.Checkout("-B", branch, remoteBranch))
 	if err != nil {
-		fmt.Println(out)
-		return fmt.Errorf("checkout remote-tracking branch %s: %w", remoteBranch, err)
+		return handleCheckoutError(
+			listBranches(gitCmd),
+			fetchFailedTag,
+			err,
+			"Fetching repository has failed",
+			branch,
+		)
 	}
 
 	return nil
--- a/gitclone/checkout_helper.go
+++ b/gitclone/checkout_helper.go
@@ -99,31 +99,21 @@ func checkoutWithCustomRetry(gitCmd git.Git, arg string, retry fallbackRetry) er
 	return nil
 }
 
-func fetchInitialBranch(gitCmd git.Git, remote string, branchRef string, fetchTraits fetchOptions) error {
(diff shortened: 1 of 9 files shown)


gitclone/checkout_helper.go
```diff
@@ -99,31 +99,21 @@ func checkoutWithCustomRetry(gitCmd git.Git, arg string, retry fallbackRetry) er
 	return nil
 }
 
-func fetchInitialBranch(gitCmd git.Git, remote string, branchRef string, fetchTraits fetchOptions) error {
```
ofalvai, line 102: Naming is hard 😵‍💫

gitclone/checkout_helper.go
```diff
@@ -99,31 +99,21 @@ func checkoutWithCustomRetry(gitCmd git.Git, arg string, retry fallbackRetry) er
...
+func forceCheckoutRemoteBranch(gitCmd git.Git, remote string, branchRef string, fetchTraits fetchOptions) error {
 	branch := strings.TrimPrefix(branchRef, refsHeadsPrefix)
 	if err := fetch(gitCmd, remote, branchRef, fetchTraits); err != nil {
-		wErr := fmt.Errorf("failed to fetch branch (%s): %w", branchRef, err)
+		wErr := fmt.Errorf("fetch branch %s: %w", branchRef, err)
 		return fmt.Errorf("%v: %w", wErr, errors.New("please make sure the branch still exists"))
 	}
-
-	if err := checkoutWithCustomRetry(gitCmd, branch, nil); err != nil {
-		return handleCheckoutError(
```
ofalvai, line 110: The custom retry here adds nicer error message (a list of existing branch names), but if a branch doesn't exist, I think it fails in the fetch step above.
I don't have a strong opinion on this, but I don't like this abstraction and I can easily say goodbye to this.

gitclone/checkout_helper.go
```diff
@@ -99,31 +99,21 @@ func checkoutWithCustomRetry(gitCmd git.Git, arg string, retry fallbackRetry) er
...
-	if err := runner.Run(gitCmd.Merge(remoteBranch)); err != nil {
-		return newStepError(
-			"update_branch_failed",
-			fmt.Errorf("updating branch (%s) failed: %w", branch, err),
-			"Updating branch failed",
-		)
+	// -B: create the branch if it doesn't exist, reset if it does
+	// The latter is important in persistent environments because shallow-fetching only fetches 1 commit,
+	// so the next run would see unrelated histories after shallow-fetching another single commit.
+	out, err := runner.RunForOutput(command.New("git", "checkout", "-B", branch, remoteBranch))
```
godrei, line 113: Shouldn't we use `gitCmd.Checkout("-B", branch, remoteBranch)` instead of a raw command? The reason is that gitCmd is initialised with a workadir, the command here only works if pwd = repo root.

(gitCmd.Checkout accepts a single argumentum, we might update to a variadic function)

gitclone/checkout_helper.go
godrei: I have vague memories about this error handling, I think the project scanner uses it. Probably the returned branches are propagated to the UI.
I suggest checking if it is in use by the scanner, or leaving it here.

ofalvai: "Unfortunately" I found references to it even in the frontend code, so this is still used: https://github.com/bitrise-io/bitrise-website/blob/2815eaa08f3d1d31bd782535dffa56c69a549dae/app/javascript/pages/AddNewApp/stages/ProjectConfig/ProjectConfig.types.ts#L11

ofalvai: I reverted `handleCheckoutError` so the above should keep working.
`checkoutWithCustomRetry` is still removed, but because it was called with a `nil` retry function, it wasn't doing anything interesting.

godrei (approved):

PR #238, opened 2026-01-30
[Step-2142] Migrate git command

tothszabi:
### Checklist

### Version

Requires a *MINOR* [version update](https://semver.org/)

### Context

TBD


gitclone/command_runner.go
```diff
@@ -46,14 +47,16 @@ func (r *DefaultRunner) RunForOutput(c *command.Model) (string, error) {
...
+	perfEnv := r.performanceMonitoringEnvVar()
+	c := t.Create(os.Stdout, io.MultiWriter(os.Stderr, &buffer), []string{perfEnv})
 
-	err := c.SetStdout(os.Stdout).SetStderr(io.MultiWriter(os.Stderr, &buffer)).Run()
+	fmt.Println()
+	log.Infof("$ %s", c.PrintableCommandArgs())
+
+	err := c.Run()
 	if err != nil {
 		if errorutil.IsExitStatusError(err) {
```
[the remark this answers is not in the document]
tothszabi: Good idea. I have done that.

gitclone/checkout_helper.go
```diff
@@ -67,11 +66,11 @@ func fetch(gitCmd git.Git, remote string, ref string, options fetchOptions) erro
 		branch = strings.TrimPrefix(ref, refsHeadsPrefix)
 	}
 
-	if err := runner.RunWithRetry(func() *command.Model {
-		return gitCmd.Fetch(opts...)
+	if err := runner.RunWithRetry(func() git.Template {
```
godrei, line 69: What is this `git.Template` ?

vendor/github.com/bitrise-io/go-utils/v2/git/template.go
```diff
@@ -0,0 +1,35 @@
...
+
+type template struct {
+	cmdFactory command.Factory
+	args       []string
+	envs       []string
+	dir        string
+}
+
+// Create ...
+func (t *template) Create(stdOut, stdErr io.Writer, envs []string) command.Command {
```
godrei, line 22: Does the command envs include `os.Environ()` too?

gitclone/command_runner.go
```diff
@@ -31,51 +31,59 @@ type DefaultRunner struct {
...
 
 	return out, err
 }
 
 // Run ...
-func (r *DefaultRunner) Run(c *command.Model) error {
+func (r *DefaultRunner) Run(t git.Template) error {
+	var buffer bytes.Buffer
+
+	perfEnv := r.performanceMonitoringEnvVar()
```
godrei: If perf monitoring is off it retruns an empty string, that we pass as envs.

gitclone/git.go
```diff
@@ -210,10 +211,11 @@ func handleCheckoutError(callback getAvailableBranches, tag string, err error, s
 	)
 }
 
-func isWorkingTreeClean(gitCmd git.Git) (bool, error) {
+func isWorkingTreeClean(gitFactory git.Factory) (bool, error) {
 	// Despite the flag name, `--porcelain` is the plumbing format to use in scripts:
 	// https://git-scm.com/docs/git-status#Documentation/git-status.txt---porcelainltversiongt
-	out, err := gitCmd.Status("--porcelain").RunAndReturnTrimmedOutput()
+	c := gitFactory.Status("--porcelain").Create(os.Stdout, os.Stderr, nil)
```
godrei: stdio set + RunAndReturnTrimmedOutput() -> this can lead to failure I think

tothszabi: You are right. Fixed it.

gitclone/command_runner.go
tothszabi: It was not a problem because the Go runtime dedups the envs before execution and also removes empty strings. But I improved the code to not to do that.

vendor/github.com/bitrise-io/go-utils/v2/git/template.go
tothszabi: Yes. The full list the git process receives is this:
```
os.Environ()  +  t.envs  +  [git perf env]
```

gitclone/checkout_helper.go
tothszabi: It is a wrapper around a command factory and some other parameters. It was needed because of how we retry the git commands. We set them up once and reexecute them when a retry is needed. This was fine with the old command package but the new one is not compatible with it because we cannot access the internal command parameters. So this wrapper save the parameters so when you call create on it it already knows what to include for the command creation.

godrei (approved):

PR #250, opened 2026-05-18
Migrate to shared steps-check workflow

ofalvai:
### Checklist

### Version

Requires a *NO* [version update](https://semver.org/)

### Context

Migrate the `check` workflow to use the shared `steps-check` repository via the `include:` directive, so the step benefits from centralized check workflow maintenance.

### Changes

- Replace the inline `check` workflow with an `include:` block pointing to `steps-check/steps.bitrise.yml`
- Fix `staticcheck` ST1005: lowercase error strings in `main.go` and `step/step.go`
- Fix `unparam`: mark intentionally unused parameters with `_` in `steperror.go` and `checkout_method_pr_auto_merge_branch.go`
- Add `//nolint:staticcheck` for intentionally capitalized "Bitrise" product name in error strings in `bitriseapi/`

### DIFF
--- a/gitclone/steperror.go
+++ b/gitclone/steperror.go
@@ -36,7 +36,7 @@ func newStepError(tag string, err error, shortMsg string) error {
 	return step.NewError("git-clone", tag, err, shortMsg)
 }
 
-func newStepErrorWithBranchRecommendations(tag string, err error, shortMsg, _ string, availableBranches []string) error {
+func newStepErrorWithBranchRecommendations(tag string, err error, shortMsg string, availableBranches []string) error {
 	// First: Map the error messages
 	newErr := newStepError(tag, err, shortMsg)
 
--- a/gitclone/steperror.go
+++ b/gitclone/steperror.go
@@ -36,7 +36,7 @@ func newStepError(tag string, err error, shortMsg string) error {
 	return step.NewError("git-clone", tag, err, shortMsg)
 }
 
-func newStepErrorWithBranchRecommendations(tag string, err error, shortMsg, currentBranch string, availableBranches []string) error {
+func newStepErrorWithBranchRecommendations(tag string, err error, shortMsg, _ string, availableBranches []string) error {
 	// First: Map the error messages
 	newErr := newStepError(tag, err, shortMsg)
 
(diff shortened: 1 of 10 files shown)


main.go
```diff
@@ -20,33 +20,33 @@ func main() {
...
 
-func run() ExitCode {
+func run() exitcode.ExitCode {
 	logger := log.NewLogger()
 	gitCloneStep := createStep(logger)
 
 	cfg, err := gitCloneStep.ProcessConfig()
 	if err != nil {
 		logger.Println()
-		logger.Errorf("%s", errorutil.FormattedError(fmt.Errorf("Failed to process Step inputs: %w", err)))
```
lpusok, line 30: In this case we may want uppercase in the error, as it is a top-level one. (Or should we change `errorutil.FormattedError` to uppercase it automatically?)

gitclone/steperror.go
```diff
@@ -36,7 +36,7 @@ func newStepError(tag string, err error, shortMsg string) error {
 	return step.NewError("git-clone", tag, err, shortMsg)
 }
 
-func newStepErrorWithBranchRecommendations(tag string, err error, shortMsg, currentBranch string, availableBranches []string) error {
```
lpusok, line 39: currentBranch param may be removed, as it seems really unused.

step/step.go
```diff
@@ -74,7 +74,7 @@ func NewGitCloneStep(logger log.Logger, tracker tracker.StepTracker, inputParser
 func (g GitCloneStep) ProcessConfig() (Config, error) {
 	var input Input
 	if err := g.inputParser.Parse(&input); err != nil {
-		return Config{}, fmt.Errorf("Error: %s\n", err)
```
lpusok, line 77: we could just return err, without appending Error (`"failed to process Step inputs` will be appended anyway)

gitclone/steperror.go
ofalvai: Thank you, fixed! (All other instances had to be aliased to `_` because the function type needs to conform to an interface.

step/step.go
ofalvai: Done!

main.go
ofalvai: I decided to revert it and add a `//nolint` directive. I don't want to bloat the current PR, we could refactor or rething that `errorutil` later

lpusok (approved):

